Issue #170, opened 2014-12-12
serviceProxy attribute IP : conceptual bug 


vlk32:
I've spent 2 days with strange behaviour. 

```
org.springframework.context.ApplicationContextException: Failed to start bean 'router'; nested exception is java.lang.RuntimeException: com.predic8.membrane.core.transport.PortOccupiedException: Opening a serversocket at port 80 failed. Please make sure that the port is not occupied by a different program or change your rule configuration to select another one.
        at org.springframework.context.support.DefaultLifecycleProcessor.doStart(DefaultLifecycleProcessor.java:170) ~[mdf-rbp-sso-2.0.0.0-SNAPSHOT.jar:na]
        at org.springframework.context.support.DefaultLifecycleProcessor.access$200(DefaultLifecycleProcessor.java:51) ~[mdf-rbp-sso-2.0.0.0-SNAPSHOT.jar:na]
        at org.springframework.context.support.DefaultLifecycleProcessor$LifecycleGroup.start(DefaultLifecycleProcessor.java:339) ~[mdf-rbp-sso-2.0.0.0-SNAPSHOT.jar:na]
        at org.springframework.context.support.DefaultLifecycleProcessor.startBeans(DefaultLifecycleProcessor.java:143) ~[mdf-rbp-sso-2.0.0.0-SNAPSHOT.jar:na]
        at org.springframework.context.support.DefaultLifecycleProcessor.start(DefaultLifecycleProcessor.java:89) ~[mdf-rbp-sso-2.0.0.0-SNAPSHOT.jar:na]
        at org.springframework.context.support.AbstractApplicationContext.start(AbstractApplicationContext.java:1278) ~[mdf-rbp-sso-2.0.0.0-SNAPSHOT.jar:na]
        at com.predic8.membrane.core.HotDeploymentThread.reload(HotDeploymentThread.java:109) ~[mdf-rbp-sso-2.0.0.0-SNAPSHOT.jar:na]
        at com.predic8.membrane.core.HotDeploymentThread.run(HotDeploymentThread.java:83) ~[mdf-rbp-sso-2.0.0.0-SNAPSHOT.jar:na]
Caused by: java.lang.RuntimeException: com.predic8.membrane.core.transport.PortOccupiedException: Opening a serversocket at port 80 failed. Please make sure that the port is not occupied by a different program or change your rule configuration to select another one.
[...]

PR #171, opened 2014-12-15
Partial fix for ISSUE #170

vlk32:
```
  <serviceProxy name="A" port="80">...</serviceProxy>
  <serviceProxy name="B" port="80" ip="localhost">...</serviceProxy>
  <serviceProxy name="C" port="80" ip="127.0.0.1">...</serviceProxy>
```
- A vs B => no solution, only runtime exception with correct error message
- B vs C => fixed

Also added :
- check if service proxy use the same SSL configuration on the same "ip:port"
- test dependencies on Mockito and PowerMock
- junit test for HttpTransport


rrayst:
Effectively merged (with adjustments due to in-between changes on master branch).

PR #172, opened 2014-12-15
Moving from log4j to slf4j + logback

vlk32:
First step of implementation of Issue Request https://github.com/membrane/service-proxy/issues/167


rrayst:
The code was migrated to slf4j a while ago. Thanks for the pull request and pushing development forward!

Issue #174, opened 2014-12-17
Stacktrace should be written to Log, not to System.err


vlk32:
As a membrane server administrator I want 
* don't print any stacktrace to `System.out` / `System.err`
  * because in server environment the standard out/err are redirected to file without log-rotate.
* use logger for all stracktraces
  * because I need redirect all messages to central log server ( via syslog ).

Now the code contains a lot of `exception.printStackTrace()` which always print exception to `System.err`.

PR #176, opened 2014-12-24
Replace e.printStackTrace() with log.warn(e.getMessage(), e)

vlk32:
Fix for #174


rrayst:
By now, this is already fixed in master.

PR #193, opened 2015-06-16
Fix the host:port extraction algorithm bug causing the <rewrite> tag failure

vweng:
Extracting host/port using double-slash instead of comparing the pathQuery with URL as the pathQuery could be decoded, which doesn't match the original URL and resulted in null schemaHost.


vweng:
Maybe I didn't explain it clear enough, this bug can easily be reproduced if the incoming URL contains any characters that were URL encoded. Because the URLUtil.getPathQuery() returns a decoded string, using dest.lastIndexOf(pathQuery) will always result in NOT FOUND, and therefore failed to retrieve the host and port from URL.

If you guys don't like to use "//" to extract the host:port, at least fix this bug by assign the url string to a new URL object and use URL.getHost() and URL.getPort() instead. Please, this is really a major bug needs to be addressed. Thanks...

PR #198, opened 2015-07-08
added cross-origin header to make the JSON response available to cross-domain API calls

vweng:
It's vital when using membrane to front JSON API used in cross domain situations.

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/interceptor/schemavalidation/JSONValidator.java
+++ b/core/src/main/java/com/predic8/membrane/core/interceptor/schemavalidation/JSONValidator.java
@@ -106,20 +106,22 @@ public class JSONValidator implements IValidator {
         for (String message : errors)
         	jg.writeString(message);
         jg.close();
-        	
+        				
 		if (failureHandler != null) {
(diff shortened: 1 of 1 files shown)


vweng:
The problem is the groovy interceptor won't work as the response is generated in the validators and will ignore everything outside. Am I right?

rrayst:
While I agree with @predic8 , your point, @vweng , is valid.

This is due to the fact that ``GroovyInterceptor`` does not override ``handleAbort()`` while the validators handle a validation error using an Outcome.ABORT, which triggers entering the error chain.

An alternative would be to move the CORS headers into a separate, new interceptor. (As we have other projects - not using validation - which would benefit from a unified CORS handling.) The new interceptor could override ``handleAbort()``.

In the mean time, enabling the groovy interceptor to handle the ABORT case in a backward compatible manner could be realized by nesting and introducing ``<abort>``:

```
<serviceProxy>
  <abort>
    <groovy>
       ...
    </groovy>
  </abort>
  <validator .../>
  ...
</serviceProxy>
```
The ``AbortInterceptor`` could be realized as ``extends AbstractFlowInterceptor`` by overriding ``handleAbort()`` and calling ``handleResponse()`` on all its children.

PR #214, opened 2015-09-20
LoadBalancingInterceptor: added missing call to DispatchingStrategy.done()

fabiankessler:
as suggested by Thomas Bayer on the mailing list https://mail.google.com/mail/u/0/#inbox/146db8f4a04962ee

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/interceptor/balancer/LoadBalancingInterceptor.java
+++ b/core/src/main/java/com/predic8/membrane/core/interceptor/balancer/LoadBalancingInterceptor.java
@@ -90,6 +90,7 @@ public class LoadBalancingInterceptor extends AbstractInterceptor {
 		}
 
 		updateDispatchedNode(exc);
+		strategy.done(exc);
 
 		return Outcome.CONTINUE;
 	}


rrayst:
All pull requests look very good so far, thank you! Looking forward to more. I'll merge them tomorrow, if fate doesn't intervene.. ;)

fabiankessler:
Hello Tobias, 

Yes that would be nice if you can merge them. All pull requests of today should go through smoothly. But if there's anything you don't like, no problem, just mention it.

Then I'll be able to update my clone from your master, and finally suggest my FaultMonitoringStrategy as an alternative to the RoundRobinStrategy.

Greetings,
Fabian

rrayst:
Hi Fabian,
I merged the request, as this makes the code a bit better than before.

Please note, that this does not solve the following two problems:
* ``LoadBalancingInterceptor`` does not implement ``handleAbort()`` (which gets called, - simply put - when an exception happens during any further ``Exchange`` handling)
* As pointed out in the Google Groups thread you pointed to, supposing a stategy decided to try nodes 1, 2 and 3, in that order, and node 1 and 2 failed, but 3 responded successfully, the strategy currently has no way of knowing the reason why nodes 1 or 2 failed (might be connection refused, might be other IOException, might be status code 5xx).<br/>
One might therefore think that the strategy is lacking relevant information when it should decide whether to auto-shutdown node 1 or 2.

Unfortunately, this is not easy to fix.

One option would be to somehow call-back the strategy from the HttpClient.

Another option would be to amend http://membrane-soa.org/service-proxy-doc/4.1/interceptors.htm to allow transitions from any of the ``handleResponse()``s and ``handleAbort()``s to the ``handleRequest()`` in the row below. Unfortunately though, I do not see a backward-compatible way of achieving this.

Best, Tobias

PR #217, opened 2015-09-21
FaultMonitoringStrategy with tests. 

fabiankessler:
How it works is documented extensively in the class header.
I believe that this implementation works well for my use case, and I've tried to make it fit for others too, but after all I can't foresee all the needs. I'm totally open to discussion.

The only existing class that is modified is HttpClient. There is some logical reordering of lines which does not change behavior. The thing that is new is the HttpClientStatusEventBus that gets informed about what is going on.


rrayst:
This has effectively been merged into master. Singletons have been refactored into Membrane's usual life cycle.

Issue #220, opened 2015-10-15
New feature: ShutdownInterceptor enables shutdown the router via HTTP-request


vlk32:
This is request for pull request #165 ( which I closed by mistake :-( ).

Hello all, 

I would like to add new feature. ShutdownInterceptor enables shutdown router via HTTP-request.
Shutdown request could (and should) be secured in the same way as others requests.

see cli/examples/shutdown
```xml
<router>
        <serviceProxy name="predic8.com" port="2000" ip="127.0.0.1">
            <!-- Accessible only via localhost interface ( ip="127.0.0.1" ).
                 Be careful, if you set up shutdown interceptor to public IP-address.
            -->
            <!-- Only for specific path.
                 Be careful, putting secret code to URL is not good idea, URLs are usually logged. -->
            <path>/shutdown</path>
            
            <!-- !!! don't forget set up security !!! This example is without security. -->

            <!-- The simplest case : any HTTP method (GET/POST/PUT/HEAD/DELETE/TRACE/CONNECT/OPTIONS) to shutdown.
            <shutdown />
            -->

            <!-- only PUT method, other methods (GET/POST/HEAD/DELETE/TRACE/CONNECT/OPTIONS) are ignored
            <shutdown httpMethod="PUT"/>
            -->

            <!-- only POST method where body contains "'secret code'" ( bodyContent is valid only for POST and PUT) -->
            <shutdown httpMethod="POST" bodyContent="'secret code'"/>

            <!-- only DELETE method, bodyContent is not significant (bodyContent is valid only for POST and PUT )
            <shutdown httpMethod="DELETE" bodyContent="anything, because it's not validated"/>
            -->
        </serviceProxy>
    </router>
```

PR #221, opened 2015-10-15
ShutdownInterceptor : Implementation for #220 ( previously discussed in #165 )

vlk32:
I've rebased old pull request #165 to the actual master.
Changeset 270926d contains thread-safe changes for this ShutdownInterceptor as was requested by  rrayst commented on 25 Dec 2014 ( https://github.com/membrane/service-proxy/pull/165#issuecomment-68101713 )


rrayst:
This has effectively just been merged into master. (with some changes to support ``<router hotDeploy="false">``, where no hot deployment thread exists)

PR #223, opened 2015-10-29
Bypass SSL cert validation when ignoreTimestampFailure flag is set to true.

vweng:
Previously setting ignoreTimestampFailure to true only ignores timestamp failure of the cert, which is of very limited use. Because if user doesn't care the expiration of cert, then most likely a self-signed cert will be used instead. In such case, bypassing the whole cert validation is often required instead of only ignoring the timestamp.


rrayst:
Please add another flag for this purpose. Redefining "ignoreTimestampFailure"? Srsly? ;)

vweng:
Urrr, don't know how to define new flags. Still pretty new to membrane... Do you mind demo some code or show me where to find an example? thx

rrayst:
Not verifying certificates presented by the peer is a security problem.

PR #226, opened 2015-11-11
Xml2json update

xorpherion:
xml2sjon update.
Updated the xml2json.xsl to the xsl from https://github.com/bramstein/xsltjson/ which has more functionality. Because it uses xsl 2.0 i had to use the saxon transformer instead of the default one. Saxon also seems to be faster.


predic8:
Can we use the saxon HE edition in our project? There are quite many people that use Membrane for their commercial products?

xorpherion:
I used saxon HE which is mozilla public license 1 and should be usable for commercial projects

PR #250, opened 2016-04-16
Maven Plugin for Service Proxy

rydnr:
We wanted to run service-proxy in our Maven project, and decided to write this initial plugin.
It works fine for us so far. We run it in isolation from the rest of the plugins.
Please review it and suggest anything you might find inconvenient.

Thank you!

### DIFF
--- a/maven-plugin/README.md
+++ b/maven-plugin/README.md
@@ -1,0 +1,32 @@
+Membrane Service Proxy Maven Plugin
+===================================
+
+This Maven plugin allows running Service Proxy from Maven.
+
+Configuration
+-------------
+
+Add the following to the `<build>` section of your *pom.xml*:
+
+```xml
+<plugin>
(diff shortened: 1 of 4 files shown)


predic8:
Hi @rydnr,
great contribution. The maven plugin will help me a lot for testing. I am wondering if we should include the plugin into Membrane core or give it a own repository as it is isolated and only depends on Membrane core. We can create the repo or you can do it also on github and we link from the documentation to the repo. It would be great if we can use some lines from your README for that.

rrayst:
The advantage of merging it into the main repo is that we can detect breaking API changes more easily concerning the plugin. (We haven't declared what is public API, and we frequently break some corner during refactoring.)

predic8:
For that we have continuous Integration

Von meinem iPhone gesendet

rrayst:
This effectively has been merged in August of 2016.

PR #289, opened 2017-09-13
remove query parameters from calledURI

erkanerol:
Problem:

In some paths, there are query parameters (e.g. GET-/pet/findByStatus in http://petstore.swagger.io/v2/swagger.json). The method "pathTemplateMatch" doesn't remove those parameters and returns false so the RuleManager cannot found a rule. The system gives "Bad Request" errors.

Solution:

The query parameters are removed before calling the Pattern.matches method.


predic8:
I am not sure if we should chop the querystring for routing. For REST it is fine but for general HTTP?

A solution to the problem is a regex path where the Querystring will be ignored. See:

https://www.membrane-soa.org/service-proxy-doc/4.4/configuration/reference/path.htm

PR #291, opened 2017-10-03
Changes to allow compilation for JDK1.6 target

FDN73:
Hello,
these  small changes allow to use membrane with target jdk 1.6 (useful for legacy hardware supporting only Red Hat 4, which can run jdk 1.6 at most) - it would be nice to keep everything jdk1.6 compatible like spring boot tries to do.


rrayst:
Membrane has targeted Java 1.7 since 2014, 1.8 since 2016. Java Language backward compatibility is not a goal of this project. Please feel free to submit patches adding special distribution variants targeting previous JVMs by auto-backporting (eg. stream-support/retro-lambda) or support for legacy hardware by conversion to native code.

PR #302, opened 2017-12-13
Fixing catchblock catching wrong exception

DJGummikuh:
Catch Block catches FileNotFoundException while in reality, it's already wrapped into a ResourceRetrievalException, which causes this catch clause to miss, allowing the Exception to travel up the entire stack and being printed on the STDOUT.


DJGummikuh:
Any chance for this to get included?

DJGummikuh:
Could you please merge that in?

rrayst:
thanks for the ping

DJGummikuh:
yay \o/ :-) Thanks!

PR #330, opened 2018-10-22
Delaying start of OAuth2 Authenticator Services

DJGummikuh:
By delaying the startup of these services selectively, I can run both the OpenID Connect Provider AND the OpenID Connect Producer in the same membrane instance, simplifying deployment of small-scale on-premise authenticated projects. Without this pull request, Membrane will fail to start

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/Router.java
+++ b/core/src/main/java/com/predic8/membrane/core/Router.java
@@ -25,6 +25,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
 import com.predic8.membrane.annot.bean.MCUtil;
+import com.predic8.membrane.core.interceptor.oauth2.authorizationservice.MembraneAuthorizationService;
 import com.predic8.membrane.core.jmx.JmxExporter;
 import com.predic8.membrane.core.jmx.JmxRouter;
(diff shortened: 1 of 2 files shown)


xorpherion:
Hi @DJGummikuh,
you can already do this by using attributes on the router element in the xml.

```
<router retryInit="true" retryInitInterval="1000">
[...]
```

tries to initialize the router and if it fails, e.g. when a rule tries to connect to another rule on startup it will postpone the failed service proxy. After retryInitInterval the router tries to start up the postponed rules. This would solve your problem with using oauth2server and oauth2resource in the same router.

Does this solve your problem?

DJGummikuh:
Hey @xorpherion!

Thanks for the quick reply. Will try that out today (its a bit wiggly since our current configuration for membrane essentially requires another hack we introduced but won't make a pull request for because it's - a hack ^^)
But besides the point wouldn't it make sense to allow Membrane to start up in such a configuration regardless of me defining retryInit (which obviously sounds like it was made for other purposes) explicitly in the configuration file? 

I see no scenario in which it is desirable for membrane to run into startup trouble that would justify to not either make the retryInit true by default OR merging my pull request which lazyloads the affected services.

Kind regards,
Johannes

rrayst:
duplicated by #364 .

PR #361, opened 2020-10-22
Issue 358 enhancement enable relative paths for chaining

precoder:


rrayst:
Using ``url.substring(0, url.indexOf("/oauth2/auth"))`` and ``contains("/oauth2/auth")`` leads to weird things e.g. when ``https://example.predic8.de/oauth2/userInfo?ignore=/oauth2/auth`` is requested.

Both patterns have therefore been replaced in 355293e1 by exact variants using ``url.startsWith(basePath + "/oauth2/auth")`` and ``basePath`` where basepath is retrieved from the surrounding ``<serviceProxy>``.

The construction
```
<serviceProxy port="80">
  <path>/abc</path>
  <oauth2AuthorizationServer ...>
```
therefore now works. In d0727c052, a test was added verifying exactly that.

PR #362, opened 2020-10-22
Issue 357 fix startup whitespace

precoder:


core/src/main/java/com/predic8/membrane/core/resolver/FileSchemaResolver.java
```diff
@@ -70,8 +71,8 @@ public void run() {
 
 	public InputStream resolve(String url) throws ResourceRetrievalException {
 		try {
-			return new FileInputStream(new File(normalize(url)));
-		} catch (FileNotFoundException e) {
+			return new FileInputStream(new File(new URI(url)));
```
rrayst, line 74: As this works only for absolute paths, it broke tests. Therefore, I went back to the ``normalize()`` method, but adjusted it in 11878ed to handle spaces correctly.

precoder: Oh yes you are right, we are generating absolute parts always but your fix is better.  😃

PR #363, opened 2020-10-22
Issue 359 enhancement all flows deliver id token

precoder:


rrayst:
As the OAuth2 specification does not issue refresh tokens in the credentials-flow and the OIDC specification, which brings in id-tokens, does not handle the credentials-flow and password-flow, we deactivated the behavior introduced by this PR by default in 869aa39a , but introduced two options to activate it:

Please use ``<oauth2authserver ``...`` issueNonSpecIdTokens="true" issueNonSpecRefreshTokens="true">``, if you want this behavior.

rrayst:
To fix the tests and to provide backward-compatibility, I removed the ``@Required`` from ``Client.setGrantTypes()`` and set the default value of ``grantTypes`` to ``authorization_code,password,client_credentials,refresh_token`` in e6693f92 .

precoder:
Thanks for your fast response!

PR #364, opened 2020-10-22
Issue 360 enhancement delayed start

precoder:


rrayst:
While we want to solve the general problem this PR tackles, the approach here has IMHO the following drawbacks:

Running the *delayed initialization* after the TCP ports opened is not thread safe. For example, the ``init2()`` method created by this PR modifies fields like ``publicAuthorizationEndpoint`` without synchronization. HTTP requests may already being processed, which causes this field to be read before/during/after its modification.

*Synchronizing access to these fields* would be an option, but delegates this (rather non-trivial) task to the developers of interceptors or nested components as ``MembraneAuthorizationService``, leading to more complicated code.

---

The current workaround we ourselves currently use is to set ``<router retryInit="true" retryInitInterval="10000">``. This causes the initialization of some ``serviceProxy``s to fail at first (becoming ``inactive`` = not handling incoming requests), but then to succeed on the second (asynchronous) attempt. Once fully initialized, the ``serviceProxy``s are thread-safely activated.

Probably we'll propose a patch in the coming days.

rrayst:
> The current workaround we ourselves currently use is to set `<router retryInit="true" retryInitInterval="10000">`. This causes the initialization of some `serviceProxy`s to fail at first (becoming `inactive` = not handling incoming requests), but then to succeed on the second (asynchronous) attempt. Once fully initialized, the `serviceProxy`s are thread-safely activated.
> 
> Probably we'll propose a patch in the coming days.

Using Membrane Service Proxy 4.8.6, another better workaround became possible:

* When the OAuth2 Authorization Server is publicly reachable below https://the-oauth2-server.example.com/server/ (e.g. https://the-oauth2-server.example.com/server/.well-known/openid-configuration in case it supports OIDC Discovery) and internally reachable on http://localhost:8001/server/ , one could always specify two (space-separated) URLs in the `<membrane src="..." />` element, the first one being the public URL (which will be used e.g. to construct `redirectUri`s during the OAuth2 flows) and the second one being the internal URL (which will be used e.g. for OIDC discovery).

* As `<internalProxy name="demo" />`s have been introduced, it should now be possible to specify `service:demo/some-path` in any `location` Membrane calls. The call will be routed to the corresponding internalProxy. While using this feature is not strictly necessary in the example, can be very helpful: having a fully-fledged proxy allows using the standard ways of configuring/manipulating HTTP requests (e.g. rewriting, adding HTTP headers, configuring TLS, ...)

Combining both features, it is possible to specify a `service:` URL as the second (=internal) URL.

The example has been confirmed. ;) The original workaround (`<router retryInit="true" retryInitInterval="10000">`) is not required for Membrane to start up correctly anymore. You possibly may want to keep it around for other reasons, though.

```xml
<serviceProxy port="443" host="the-oauth2-server.example.com">
	<spring:ref bean="ssl" />
	<target host="localhost" port="8001" />
</serviceProxy>

<internalProxy name="as">
	<target host="localhost" port="8001" />
</internalProxy>

<serviceProxy port="443" host="some-oauth2-client.example.com">
	<spring:ref bean="ssl" />
	<oauth2Resource2>
		<jwtSessionManager>
			...
		</jwtSessionManager>
		<membrane src="https://the-oauth2-server.example.com service:as/server" clientId="..." clientSecret="..." scope="openid profile" />
	</oauth2Resource2>
	<target host="localhost" port="8002" />
</serviceProxy>
```

Issue #392, opened 2022-11-14
Redirect on AuthWithSessionRequest Auth URL causes infinite redirect loop


precoder:
Hello,

If for some reason call for token revocation not works or a race condition happens on login call from the Browser side, we get an infinite redirect loop after the first login/logout cycle.

AuthWithSessionRequest clears the Session properties and calls the getRequestURI again, but since the Session is partially available AuthWithSessionRequest is called again. This causes clearing the Session properties and calling the same getRequestURI again which is an endless loop.

Instead of only clearing session, if AuthWithSessionRequest completely removes the session, next call lands into AuthWithoutSessionRequest and we can login again.

Here is the call URL and the redirect response from the Membrane:

`/auth/oauth2/auth?response_type=code&client_id=test123&redirect_uri=https://localhost:8843/api/sessionmanagement/oauth2callback&state=stateid123&scope=openid%20offline_access&prompt=login

HTTP/1.1 307 Temporary Redirect
Location: /auth/oauth2/auth?response_type=code&client_id=test123&redirect_uri=https://localhost:8843/api/sessionmanagement/oauth2callback&state=stateid123&scope=openid%20offline_access&prompt=login
Content-Type: text/html;charset=UTF-8
Content-Length: 0
Expires: Tue, 03 Jul 2001 06:00:00 GMT
Cache-Control: no-store, no-cache, must-revalidate, max-age=0
Cache-Control: post-check=0, pre-check=0
Pragma: no-cache

<html>...</html>
`

rrayst:
Good work, thanks!

PR #393, opened 2022-11-14
Fix issue #392 by removing session completely instead of only cleaning session properties.

precoder:


predic8:
@precoder Thanks for the contrib

@rrayst Looks good to me. Could you check and approve?

PR #445, opened 2023-02-02
message, header, queryparams variables for template plugin, json and …

predic8:
…xml samples, constants, refactorings


core/src/main/java/com/predic8/membrane/core/Constants.java
```diff
@@ -53,11 +51,10 @@ public class Constants {
 
 	public static final String XML_VERSION = "1.0";
 
-	public static final String UTF_8 = "UTF-8";
-	public static final Charset UTF_8_CHARSET = Charset.forName(UTF_8);
+	//public static final Charset UTF_8_CHARSET = StandardCharsets.UTF_8;
```
predic8: Delete

core/src/main/java/com/predic8/membrane/core/Constants.java
```diff
@@ -53,11 +51,10 @@ public class Constants {
...
 
 	public static final String ISO_8859_1 = "ISO-8859-1";
-	public static final Charset ISO_8859_1_CHARSET = Charset.forName(ISO_8859_1);
+	//public static final Charset ISO_8859_1_CHARSET = StandardCharsets.ISO_8859_1;
```
predic8: Delete

core/src/main/java/com/predic8/membrane/core/Constants.java
```diff
@@ -90,7 +84,7 @@ public class Constants {
 	public static final String PRODUCT_CONTACT_EMAIL = "info@predic8.de";
 
 	public static final String HTML_FOOTER =
-			"Copyright �2009-2023 " +
+			"Copyright ©2009-2023 " +
```
rrayst, line 87: Das rendert nicht richtig: 
image

core/src/main/java/com/predic8/membrane/core/exchangestore/FileExchangeStore.java
```diff
@@ -173,7 +160,7 @@ public void initializeTimer() {
 			return; // don't do anything if this feature is deactivated
 		}
 
-		oldFilesCleanupTimer = new Timer("Clean up old log files", true);
+		Timer oldFilesCleanupTimer = new Timer("Clean up old log files", true);
```
rrayst, line 163: `Timer`s have the following issues:
* each timer spawns 1 `Thread`. In case a shared Timer can be used (=the task is not too time sensitive), `router.getTimerManager()` should be preferred.
* The `Timer` here is started, but never stopped. This means that each Hot Deployment starts a new thread.

core/src/main/java/com/predic8/membrane/core/http/Header.java
```diff
@@ -106,21 +109,18 @@ public class Header {
...
 	public static final String MAX = "max";
 
 	public static final String CLOSE = "close";
 
-	private static final Pattern mediaTypePattern = Pattern.compile("(.+)/([^;]+)(;.*)?");
-	private static final Pattern parameterPattern = Pattern.compile("(.+)=\"?([^\"]+)\"?");
-
-	private static final Pattern timeoutPattern = Pattern.compile("timeout\\s*=\\s*(\\d+)", Pattern.CASE_INSENSITIVE);
-	private static final Pattern maxPattern = Pattern.compile("max\\s*=\\s*(\\d+)", Pattern.CASE_INSENSITIVE);
+	private static final Pattern mediaTypePattern = compile("(.+)/([^;]+)(;.*)?");
```
rrayst: `mediaTypePattern` und `parameterPattern` werden nicht mehr verwendet.

core/src/main/java/com/predic8/membrane/core/lang/ScriptingUtils.java
```diff
@@ -0,0 +1,70 @@
...
+        if (msg != null) {
+            parameters.put("message", msg);
+            parameters.put("header", msg.getHeader());
+            parameters.put("body", msg.getBodyAsStringDecoded());
+            if (includeJsonObject) {
+                try {
+                    log.info("Parsing body as JSON for scripting plugins");
+                    parameters.put("json",om.readValue(readInputStream(msg.getBodyAsStream()),Map.class));  // @Todo not with Javascript
+                } catch (Exception e) {
+                    log.warn("Can't parse body as JSON: " + e);
```
rrayst, line 61: This is bad practice, as it swallows the stacktrace of the exception. Please use `log.warn("Can't parse body as JSON: ", e);` instead.

rrayst (changes requested):

core/src/main/java/com/predic8/membrane/core/lang/ScriptingUtils.java
predic8: Here the stacktrace do not provide more details. All the Info about the parsing is in the message. And we get a stacktrace in the Log everytime somebody sends a request containing invalid JSON.

core/src/main/java/com/predic8/membrane/core/http/Header.java
predic8: Did you delete the right lines? Code does not match comment.

rrayst (approved):

PR #450, opened 2023-02-07
AdminConsole Searchfilter for Text

predic8:
There are some tests that fail. The tests are fixed in the other pull requests. **Process that after the other three.**


predic8:
There was a requirement also to search for Header but I do not know if this is really needed.

rrayst (approved):

rrayst (approved):

PR #463, opened 2023-02-14
Removed deprecation sortMediaTypeByQuality #458

predic8:


core/src/main/java/com/predic8/membrane/core/interceptor/rest/REST2SOAPInterceptor.java
```diff
@@ -183,21 +185,25 @@ public Outcome handleResponse(Exchange exc) throws Exception {
...
-		if (inputIsXml && wantedType >= 1) {
-			response.setBodyContent(xml2json(response.getBodyAsStreamDecoded(), properties));
-			setJSONContentType(response.getHeader());
-		}
+		response.setBodyContent(xml2json(response.getBodyAsStreamDecoded(), properties));
+		setJSONContentType(response.getHeader());
+	}
+
+	private static MediaType getMediaTypeWithHighestQualityFromAcceptHeader(Header requestHeader) {
+		return sortMimeTypeByQualityFactorAscending(requestHeader.getFirstValue(ACCEPT)).get(0);
```
rrayst: possible IndexOutOfBoundsException, if the list is empty

rrayst (changes requested):

core/src/main/java/com/predic8/membrane/core/interceptor/rest/REST2SOAPInterceptor.java
predic8: Fixed

rrayst (approved):

PR #470, opened 2023-02-18
No limit for connection count from same Ip

predic8:


core/src/main/java/com/predic8/membrane/core/transport/Transport.java
```diff
@@ -30,13 +30,18 @@
...
 	private Router router;
 	private boolean printStackTrace = false;
 	private boolean reverseDNS = true;
-	int concurrentConnectionLimitPerIp = 60;
+
+	/**
+	 * Number of allowed concurrent connections from one client ip
+	 * @default -1 No Limit
+	 */
+	private int concurrentConnectionLimitPerIp = -1;
```
rrayst, line 40: This Javadoc comment will *not* be picked up by the annotation processor, as the target is not a setter method (but a field).

test/src/main/python/connections.py
```diff
@@ -0,0 +1,14 @@
...
+# Install: pip3 install locust
+# Run: locust -f connections.py
+#
+# See: https://docs.locust.io/en/stable/index.html
+class WebsiteUser(HttpUser):
+    # wait_time = between(5, 15)
+
+    @task
+    def fruit(self):
+        self.client.get("/shop/products/")
```
rrayst: Was this file added by accident?

rrayst (changes requested):

core/src/main/java/com/predic8/membrane/core/transport/Transport.java
predic8: Ok. I'll move the comment to the setter.

test/src/main/python/connections.py
predic8: No. But it is somewhat hidden. This is a locust script to provoke the limit from above. And it is useful for loadtesting. I'll move it and add some documentation.

rrayst (approved):

Issue #476, opened 2023-02-20
Option for Error internal Details in ProblemDetails [For Discussion]


predic8:
Alternatives

<router mode="dev|prod">

<router suppressInternalErrors="yes">

See SOAPStacktraceFilter

predic8:
Default should be development

rrayst:
set `@Required` on the option for 5.1.0?

PR #477, opened 2023-02-20
Documentation fixes, typos, Javascript Plugin Error Logs

predic8:


core/src/main/java/com/predic8/membrane/core/lang/AbstractScriptInterceptor.java
```diff
@@ -69,7 +70,15 @@ protected Outcome runScript(Exchange exc, Flow flow) throws InterruptedException
...
         Message msg = getMessage(exc, flow);
 
-        Object res = script.apply(getParameterBindings(exc, flow, msg));
+        Object res;
+        try {
+            res = script.apply(getParameterBindings(exc, flow, msg));
+        } catch (Exception e) {
+            Map<String,Object> details = new HashMap<>();
+            details.put("message","See logs for details.");
+            exc.setResponse(createProblemDetails(500, "/internal-error", "Internal Server Error", details));
```
rrayst, line 79: shouldn't this be something like `createProblemDetails(500, "/scripting-error"` ?

predic8: I also thought about that. But do we want to expose internals to the outside, like the fact that there is a script?

rrayst: So it's related to #476 ...

rrayst (approved):

PR #495, opened 2023-03-03
Redundancy reduction

bridgerdier:
Selectively applied IntelliJ Suggestions to remove redundant code constructs


core/src/main/java/com/predic8/membrane/core/kubernetes/KubernetesWatcher.java
```diff
@@ -87,8 +87,8 @@ private KubernetesClient getClient() {
 
     private Optional<KubernetesValidationInterceptor> findK8sValidatingInterceptor() {
         return router.getRules().stream()
-                .filter(rule -> rule.getInterceptors() != null)
                 .map(rule -> rule.getInterceptors())
+                .filter(interceptors -> interceptors != null)
```
predic8, line 91: One line might be a duplicate:

.filter(interceptors -> interceptors != null)
.filter( i -> i != null)

predic8 (dismissed):

predic8 (approved):

PR #496, opened 2023-03-04
Fix permission error in FileUtilTest

bridgerdier:
Test attempted to create a temp file at "/tmptest.tmp", resulting in a permission error. Added path separator to temp file path


bridgerdier:
Alternatively, use java.nio.file.Files.createTempFile

predic8 (approved):
Thanks

PR #524, opened 2023-04-15
Fix for Return from Groovy Interceptor with GraalVM Value

predic8:


predic8:
There are also some minor improvements reqarding documentation and samples

rrayst (approved):

PR #542, opened 2023-05-25
A Option to verify config file and shutdown afterwards

bridgerdier:


predic8 (commented):
Can you rename the parameter to -t to make it more consistant to the others?

rrayst (approved):

PR #557, opened 2023-05-31
recursively kill all children of spawned process instead of shell magic

bridgerdier:
if it works as expected, Process2 can be significantly simplified afterwards (startCommand)


predic8 (approved):
Looks always nice when there are more deletions than adds :-)

PR #571, opened 2023-06-02
improved Embedding java example

christiangoerdes:


distribution/examples/embedding-java/pom.xml
```diff
@@ -5,7 +5,32 @@
...
     <version>1.0-SNAPSHOT</version>
 
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.1</version>
+                <configuration>
+                    <source>7</source>
```
predic8: Change to Java 8

distribution/examples/embedding-java/src/main/java/com/predic8/membrane/api/AddMyHeaderInterceptor.java
```diff
@@ -26,16 +26,11 @@
  * @author Oliver Weiler
  */
 public class AddMyHeaderInterceptor extends AbstractInterceptor {
-    private static final Logger log = Logger.getLogger(AddMyHeaderInterceptor.class);
-
     @Override
     public Outcome handleRequest(Exchange exchange) {
-        log.info(exchange.getRequest().getHeader().getUserAgent());
-
+        System.out.println(exchange.getRequest().getHeader().getUserAgent());
```
predic8: Use logger instead of print

distribution/examples/embedding-java/src/main/java/com/predic8/membrane/api/EmbeddingJava.java
```diff
@@ -40,16 +42,20 @@ public static void main(String[] args) throws Exception {
         String path = ".*";
         int listenPort = 4000;
 
-        // The API should listen to GET requests on port 4000 with any path from any host
         ServiceProxyKey key = new ServiceProxyKey(hostname, method, path, listenPort);
+        key.setMethod(method);
```
predic8, line 46: Is it needed to set the key and method when you are passing them in the constructor

predic8: Can you exchange the ServiceProxyKey with an APIProxyKey? There should be one I guess.

predic8 (changes requested):

PR #590, opened 2023-06-07
Added example for jsonProtectionInterceptor

t-burch:


distribution/examples/json-protection/requests.bat
```diff
@@ -0,0 +1,49 @@
+@echo off
+
+echo --------- Valid Request ---------
+echo.
+curl -d @requests/valid.json localhost:2000 | python -m json.tool
```
predic8: Take out python use beautifier Interceptor instead

predic8 (changes requested):

predic8 (approved):

PR #619, opened 2023-06-15
use CVN to keep RPM spec up to date

bridgerdier:
Update ConsistentVersionNumbers to handle RPM spec as well


bridgerdier:
Additional change coming up

rrayst (approved):

Issue #622, opened 2023-06-16
Json-protection should ignore new lines 


christiangoerdes:
Prettified JSON and minimized JSON have different string lengths, even though whitespace is ignored

PR #623, opened 2023-06-16
implemented JsonProtectionTest

christiangoerdes:
Note: First test not working because:  #622


predic8 (changes requested):
Refactor the test to a parametrized test:

See:
https://www.tutorialspoint.com/junit/junit_parameterized_test.htm

Specify the filename and the expected outcome:

valid.json 200
max_tokens.json 400

predic8 (approved):

predic8:
size counts the bytes send. If there are a lot of whitespaces, the protection will kick in earlier. Thats ok. The test should be go through.

PR #632, opened 2023-06-18
Codeql fixes

predic8:


bridgerdier (approved):
Does the job, but I still think counting interceptors in a stream is more explicit.

predic8:
Thanks. You are right. Just wanted the codeql check to go through.

PR #636, opened 2023-06-18
Openapi rewritter 601

predic8:


core/src/main/java/com/predic8/membrane/core/util/URI.java
rrayst, line 1: The changes do not fit to the class comment "Same behavior as java.net.URI".

rrayst (changes requested):

core/src/main/java/com/predic8/membrane/core/util/URI.java
rrayst: e.g. `getPort()` as proposed has different semantics from `java.net.URI.getPort()`.

rrayst: Using non-standard semantics on a class called `URI` might be quite confusing to most Java Developers.

rrayst (approved):

PR #640, opened 2023-06-22
Basic auth example relocation

t-burch:
Moved the basic auth example to security section


t-burch:
Commits from branch 639 included in this one again, but shouldn't break anything

predic8 (approved):

PR #668, opened 2023-07-16
Examples and test fixes

predic8:
@rrayst Ask me before reviewing and merging

- A test was duplicated and covered by other tests


rrayst:
The [tutorial](https://www.membrane-soa.org/api-gateway-doc/current/oauth2-password-flow-sample.html) says
```
"OK" will appear in the command line.
```
when running `client.ps1`, but this is not the case.

rrayst:
documentation was adjusted.

rrayst (approved):

PR #672, opened 2023-07-18
Versioning example rewrite

t-burch:
Rewrote versioning sample readmes and tested them


predic8 (approved):

distribution/examples/versioning/routing/README.md
```diff
@@ -1,66 +1,72 @@
...
- curl --header "Content-Type: text/xml" -d @request_v20.xml http://localhost:2000/ContactService
-```
-  Observe that both requests, `1.1` and `2.0`, get a response from their respective service, although
-  the endpoint used is the same.
+(Note that the WSDL files reference the actual endpoint (depending on the
+service version), and using them in any client (e.g. SoapUI) therefore
+bypasses the proxy.)
 
 ---
 See:
```
t-burch, line 72: Still missing reference url

PR #690, opened 2023-08-11
Setheader default

koin612:
I'm also thinking to change the config from

`<setHeader name="Cache-Control" value="no-store" ifAbsent="true" />`

to

`<setHeader name="Cache-Control" valueIfAbsent="no-store" />`

the second looks better to me but I feel like the first one fits better into current membrane config.

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/http/Header.java
+++ b/core/src/main/java/com/predic8/membrane/core/http/Header.java
@@ -191,6 +191,11 @@ public class Header {
 		return fields.toArray(new HeaderField[0]);
 	}
 
+	public boolean contains(String header) {
+		return fields.stream()
+				.anyMatch(headerField -> headerField.getHeaderName().equals(header));
+	}
+
 	/**
 	 * Since {@link HttpUtil#readLine(InputStream)} assembles the String byte-by-byte
 	 * converting it to char-by-char, we use ISO-8859-1 for output here.
--- a/core/src/main/java/com/predic8/membrane/core/interceptor/misc/SetHeaderInterceptor.java
+++ b/core/src/main/java/com/predic8/membrane/core/interceptor/misc/SetHeaderInterceptor.java
@@ -24,6 +24,7 @@ import org.springframework.expression.*;
 import org.springframework.expression.spel.standard.*;
 import org.springframework.expression.spel.support.*;
 
+import java.util.Arrays;
 import java.util.regex.*;
 
 import static com.predic8.membrane.core.interceptor.Outcome.*;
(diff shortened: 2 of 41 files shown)


core/src/main/java/com/predic8/membrane/core/interceptor/misc/SetHeaderInterceptor.java
```diff
@@ -72,7 +83,12 @@ public Outcome handleResponse(Exchange exc) throws Exception {
     }
 
     private Outcome handleMessage(Exchange exchange, Message msg) {
-        msg.getHeader().setValue(name, evaluateExpression(new ExchangeEvaluationContext(exchange, msg).getStandardEvaluationContext()));
+        var msgContainsHeader = Arrays.stream(msg.getHeader().getAllHeaderFields()).anyMatch(headerField -> headerField.getHeaderName().equals(name));
```
predic8: add hasField() to Header(s)
make sure case is ignored
add tests

core/src/main/java/com/predic8/membrane/core/interceptor/misc/SetHeaderInterceptor.java
```diff
@@ -72,7 +83,10 @@ public Outcome handleResponse(Exchange exc) throws Exception {
     }
 
     private Outcome handleMessage(Exchange exchange, Message msg) {
-        msg.getHeader().setValue(name, evaluateExpression(new ExchangeEvaluationContext(exchange, msg).getStandardEvaluationContext()));
+        if (!ifAbsent || !msg.getHeader().contains(name)) {
```
predic8: Can you reverse the condition and make it as a guard? It is hard to understand unterwise.

core/src/main/java/com/predic8/membrane/core/http/Header.java
```diff
@@ -191,6 +191,11 @@ public HeaderField[] getAllHeaderFields() {
 		return fields.toArray(new HeaderField[0]);
 	}
 
+	public boolean contains(String header) {
+		return fields.stream()
+				.anyMatch(headerField -> headerField.getHeaderName().equals(header));
```
predic8: Do we need ignorecase?
Add tests to HeaderTest

predic8 (changes requested):

core/src/main/java/com/predic8/membrane/core/http/Header.java
koin612: Header -> HeaderName class already handles equals ignoring case in it's equals override

core/src/main/java/com/predic8/membrane/core/interceptor/misc/SetHeaderInterceptor.java
koin612: but then its like

```java
if (bool) return CONTINUE;
setStuff()...
return CONTINUE;
```

and the function always and multiple times returns CONTINUE

koin612: changed to guard

predic8 (approved):

predic8 (approved):

PR #702, opened 2023-08-23
updated synchronization; added test

koin612:


predic8:
At first glance looks good, but Tobias should have a look at it.

rrayst (approved):

PR #726, opened 2023-09-07
wip enabled basic logging and configuring in apache commons log forma…

koin612:
Current `log4j2.xml` configuration would enable logs like:

`127.0.0.1 [07/09/2023:17:15:20 +0200] "GET / HTTP/1.1" 200 13187`

It could be configured like this:

```xml
<api port="2000">
    <accessLog>
        <additionalPattern create="port" withExchange="originalHostHeaderPort" override="false" />
        <additionalPattern create="response.status" withExchange="response?.statusCode" orDefaultValue="NOT" />
    </accessLog>
    <target host="www.predic8.de" port="80" />
</api>
```

where `additionalPattern` is optional and provides some configurability like the `accessLog`.

I'd like to discuss:

1. if there could be a use-case for request logging, if not I could simplify the MDC-Map-creation. It was first designed with request logging in mind and might be messier than needed.
2. `exchange.getResponse().getBody().getLength()` and some other methods for content payload size are commented with this warning `Warning: Calling this method will trigger reading the body from the client, disabling "streaming".`. Is this a Issue? Is there a good way around?
3. Should we log authentication credentials like email/names or something to identify the user triggering the call?

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/interceptor/log/AccessLogInterceptor.java
+++ b/core/src/main/java/com/predic8/membrane/core/interceptor/log/AccessLogInterceptor.java
@@ -1,0 +1,155 @@
+package com.predic8.membrane.core.interceptor.log;
+
+import com.predic8.membrane.annot.MCAttribute;
+import com.predic8.membrane.annot.MCChildElement;
+import com.predic8.membrane.annot.MCElement;
+import com.predic8.membrane.core.exchange.Exchange;
+import com.predic8.membrane.core.interceptor.AbstractInterceptor;
+import com.predic8.membrane.core.interceptor.Outcome;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
+import org.springframework.expression.spel.standard.SpelExpressionParser;
+
+import java.text.SimpleDateFormat;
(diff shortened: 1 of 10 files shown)


predic8 (changes requested):
Can you add an example under examples/logging/access and also add an example test?

koin612:
> Can you add an example under examples/logging/access and also add an example test?

I've done some examples and push them later on. Creating tests was harder than expected and failed for now.

- User information with status code 401 cannot be trusted
- If I prepend an `jwtAuth` interceptor and it cancels the exchange because its not valid, should the access log interceptor log this request? 

I think we should talk on Monday about these points.

predic8:
> > Can you add an example under examples/logging/access and also add an example test?
> 
> I've done some examples and push them later on. Creating tests was harder than expected and failed for now.
> 
>     * User information with status code 401 cannot be trusted
> 
>     * If I prepend an `jwtAuth` interceptor and it cancels the exchange because its not valid, should the access log interceptor log this request?

Hi @koin612 ,
look at the example tests in the distribution/src/test folder. They are pretty simple. Just start an example, send a request mit restassured and check if the logfile gets created. There many samples. Look at the class LoggingCSVTest.

predic8 (approved):

PR #749, opened 2023-09-21
fix code analysis high prio bugs and opened discussion for unknown

koin612:


core/src/main/java/com/predic8/membrane/balancer/client/LBNotificationClient.java
```diff
@@ -45,9 +43,10 @@
 
 public class LBNotificationClient {
 
-	private static Logger log = LoggerFactory.getLogger(LBNotificationClient.class.getName());
+	private static final Logger log = LoggerFactory.getLogger(LBNotificationClient.class.getName());
+	private static final SecureRandom secureRandom = new SecureRandom();
```
koin612, line 47: [Random created and used only once](https://spotbugs.readthedocs.io/en/latest/bugDescriptions.html#dmi-random-object-created-and-used-only-once-dmi-random-used-only-once)

PR #753, opened 2023-10-02
Added limit option for amount of mutations in a GraphQL query document

t-burch:


core/src/main/java/com/predic8/membrane/core/graphql/GraphQLProtectionInterceptor.java
```diff
@@ -159,6 +160,17 @@ public Outcome handleRequest(Exchange exc) throws Exception {
...
+        int mutationCount = 0;
+        for (ExecutableDefinition definition : ed.getExecutableDefinitions()) {
+            if (definition instanceof OperationDefinition)
+                if (((OperationDefinition) definition).getOperationType() != null
+                        && ((OperationDefinition) definition).getOperationType().getOperation().equals("mutation"))
+                    mutationCount++;
+        }
+
+        if (mutationCount > maxMutations)
+            error(exc, 405, "Too many mutations defined in document.");
```
predic8: I would use 400 Bad Request, 405 if for wrong HTTP Method like POST or GET

core/src/main/java/com/predic8/membrane/core/graphql/GraphQLProtectionInterceptor.java
```diff
@@ -159,6 +160,17 @@ public Outcome handleRequest(Exchange exc) throws Exception {
 
         ExecutableDocument ed = graphQLParser.parseRequest(new ByteArrayInputStream(((String) query).getBytes(UTF_8)));
 
+        int mutationCount = 0;
+        for (ExecutableDefinition definition : ed.getExecutableDefinitions()) {
+            if (definition instanceof OperationDefinition)
```
predic8: Can you write this as stream with lambdas? There should also be a test for the count.

predic8 (changes requested):

core/src/main/java/com/predic8/membrane/core/graphql/GraphQLProtectionInterceptor.java
```diff
@@ -203,6 +205,15 @@ public Outcome handleRequest(Exchange exc) throws Exception {
         return Outcome.CONTINUE;
     }
 
+    public int countMutations(List<ExecutableDefinition> definitions) {
+        return (int) definitions.stream()
+                .filter(definition -> definition instanceof OperationDefinition)
+                .map(definition -> (OperationDefinition) definition)
+                .filter(operation -> operation.getOperationType() != null)
+                .filter(operation -> operation.getOperationType().getOperation().equals("mutation"))
```
rrayst, line 213: kann `operation.getOperationType().getOperation()` null sein?

t-burch: Ohne das hatte es einen null Fehler geworfen

PR #769, opened 2023-10-24
Refactored web server

t-burch:


predic8 (approved):

rrayst:
The index page should only be generated, if `generateIndex` is set to `true`.

predic8:
Add test for generateIndex.

predic8 (approved):

PR #868, opened 2024-02-13
Migrated HTTP 1.0 & 1.1 tests to SampleSoapService

t-burch:


core/src/test/java/com/predic8/membrane/integration/Http10Test.java
```diff
@@ -72,32 +80,31 @@ public void testPost() throws Exception {
 
 		String response = post.getResponseBodyAsString();
 		assertNotNull(response);
```
predic8, line 83: Test StatusCode, Content like Munich, .., ContentType

core/src/test/java/com/predic8/membrane/integration/Http11Test.java
```diff
@@ -84,7 +90,6 @@ private void testPost(boolean useExpect100Continue) throws Exception {
 		assertEquals(200, status);
```
predic8, line 90: Asserts like HTTP 1.0

predic8 (approved):

PR #870, opened 2024-02-13
SoapProxyIntegrationTest Fix

t-burch:


predic8 (commented):
Order raus
Methoden umbennen: shutdown, test1, test2

predic8 (commented):
Make it as Integration test again, against azure proxy and the sampleSoap...

predic8 (approved):

PR #930, opened 2024-03-05
Updated membrane core dependencies in examples to membrane version 5.3.3

t-burch:


rrayst (changes requested):
can we do this automatically using `ConsistentVersionNumbers`?

t-burch:
It should already do it, not sure why it does not work

predic8 (approved):

predic8 (approved):

christiangoerdes (approved):

rrayst (approved):

Issue #1011, opened 2024-04-05
Examples: APIKey + OpenAPI Securiy Validation


predic8:
in examples/security/apikey/apikey-openapi

Sample like:

´´´
 <api port="3000">
                <openapi location="fruitshop-api-v2-1-0-security.yml" validateRequests="no" validateSecurity="yes"/>
                <apiKey required="false">
                    <headerExtractor name="X-API-KEY" />
                </apiKey>

            </api>

      security:
        - http:
          - write

  securitySchemes:
    http:
      type: apiKey
      in: header
      name: X-API-KEY

´´´

predic8:
- Take the fruitshop OpenAPI
- Require all deletes the scope delete
- Require PUT and POST write
- GET are possible just with an API-KEY: Global at top
- Make 3 Keys:
111:
222: write
333: write,delete

PR #1014, opened 2024-04-09
#1011 api-key + open api validation example

christiangoerdes:


distribution/examples/security/api-key/apikey-openapi/README.md
```diff
@@ -0,0 +1,81 @@
+# Swagger UI: API Key Authorization
+
+Enhance security using API keys with role-based access control (RBAC).
```
predic8: This example shows how to use API keys together with role-based access control (RBAC). An API key can have associated scopes(roles) that are matched against security definitions in OpenAPI documents.

The Swagger UI is used as a client that authenticates with the backend.

distribution/examples/security/api-key/apikey-openapi/README.md
```diff
@@ -0,0 +1,81 @@
+# Swagger UI: API Key Authorization
```
predic8: Authorization with API Keys and OpenAPI Security Schemes

predic8 (approved):

Issue #1012, opened 2024-04-05
OpenAPI SecurityValidator: Configuration 


predic8:
With validateSecurity="yes" alone nothing happens. To make it work validateRequests="yes" is at the same time required.

Alternatives:

a.) Activate validateRequests automatically with setting validateSecurity

b.) Do not start when validateRequests=no and show an error

c.) Add a flag security to ValidationError. If   validateSecurity is yes, validate allways requests. Look at the errors. If there is no security error let it pass. If there is a security error return error message to the caller.

predic8:
I would opt for c.) but I'am not sure. Reason: Most intuitive for the user.

t-burch:
What about an enum like:
`validationScheme="<REQUEST|RESPONSE|BOTH|SECURED_REQUEST|SECURED_BOTH>"`

PR #1015, opened 2024-04-09
Implemented #1012-SecurityValidator-Configuration Solution

christiangoerdes:


predic8:
@christiangoerdes 

If validateSecurity=YES than validateRequest should also be activated. Show a Warning in the log, that validateRequest was automatically activated.

christiangoerdes:
I already implemented this, but it looks like I forgot to push it. Maybe @t-burch can push it on Monday.

predic8 (approved):

PR #1028, opened 2024-04-18
Added examples for conditional interceptor

t-burch:


distribution/conf/proxies.xml
```diff
@@ -1,31 +1,72 @@
...
-<!--            </apiKey>-->
-            <target host="localhost" port="2001" />
-        </api>
+            <request>
+            <!--    <groovy>
+                    exc.setProperty("pi", exc.getRequest().getBodyAsStringDecoded())
+                </groovy>
+
+                <if test="request.getBody.getLength gt 64" language="SpEL">
+                    <groovy>System.out.println("Body too large!")</groovy>
```
predic8: println
only one return at the end

predic8 (approved):

PR #1040, opened 2024-04-24
Escape double quotes in accessLogInterceptor custom vars insertions.

t-burch:
Added test


core/src/main/java/com/predic8/membrane/core/interceptor/log/AccessLogInterceptorService.java
```diff
@@ -136,12 +136,16 @@ private String safe(Supplier<Object> access) {
...
     private String safe(Supplier<Object> access, String defaultValue) {
         try {
-            return String.valueOf(access.get());
+            return escapeQuotes(String.valueOf(access.get()));
         } catch (Exception e) {
             return defaultValue;
         }
     }
 
+    static String escapeQuotes(String s) {
```
predic8: Should be either private or better moved to a Utility class

predic8 (changes requested):

predic8 (approved):

Issue #1058, opened 2024-05-03
More fine grained assertions for WellknownFileTest


predic8:
Ask Thomas

PR #1070, opened 2024-05-08
Added more fine grained assertions for WellknownFileTest #1058

christiangoerdes:


rrayst (changes requested):
The test previously verified that JSON Serialization of a `WellknownFile` instance works.

Now it *only* tests getters & setters.

Can we test serialization again?

t-burch:
@rrayst Added old test back additionally

rrayst (approved):

PR #1107, opened 2024-05-27
Added util function to find free port above set threshold and changed tests to use this mechanism

t-burch:


core/src/test/java/com/predic8/membrane/integration/SessionManager.java
```diff
@@ -122,7 +128,7 @@ public void sessionExpires(
         assertNotEquals(rememberThis,rememberThisFromServer);
         assertEquals("null", rememberThisFromServer);
 
-        httpRouter.stop();
+        httpRouter.shutdown();
```
predic8, line 131: Change back

core/src/test/resources/soap-proxy.xml
```diff
@@ -6,15 +6,15 @@
 					    
 	<router>
 
-		<soapProxy port="2000" wsdl="http://localhost:3000/?wsdl"/>
+		<soapProxy port="4000" wsdl="http://localhost:3000/?wsdl"/>
 
-		<soapProxy port="2001" wsdl="http://localhost:3000/?wsdl">
+		<soapProxy port="4001" wsdl="http://localhost:3000/?wsdl">
```
predic8: -1 and ask router in test for proxy with name=...

predic8 (approved):

PR #1145, opened 2024-06-24
setProperty Interceptor

t-burch:
Added setProperty interceptor and generalized setHeader and setProperty interceptors


core/src/main/java/com/predic8/membrane/core/interceptor/misc/SetHeaderInterceptor.java
```diff
@@ -8,16 +8,20 @@
...
     }
 
     @Override
-    protected void setValue(Exchange ignored, Message msg, String eval) {
-        msg.getHeader().setValue(name, eval);
+    protected void setValue(Exchange exc, Flow flow, String eval) {
+        msgFromExchange(exc).getHeader().setValue(name, eval);
+    }
+
+    private Message msgFromExchange(Exchange exc) {
```
predic8: Should be dependent on direction

predic8 (changes requested):

predic8 (approved):

PR #1151, opened 2024-06-26
Migrated RuleKey complexMatch and surrounding methods to use Exchange

t-burch:


predic8:
@rrayst Can you have a look at the changes. I think we can change the interface. We need that to make the matching more powerful like:

<api port="2000" test="param.customer == 7"...>

PR #1169, opened 2024-07-03
added dynamic rewriter for openApiDocs

christiangoerdes:


core/src/main/java/com/predic8/membrane/core/util/Util.java
```diff
@@ -116,4 +118,13 @@ public static ExecutorService createNewThreadPool() {
 	public static List<String> splitStringByComma(String string) {
 		return new ArrayList<>(Arrays.asList(string.split(",")));
 	}
+
```
predic8: - Apidoc comment
- Junit tests

predic8 (dismissed):

predic8 (approved):

PR #1179, opened 2024-07-08
Added request & response debugging for AuthenticationApi

t-burch:


core/src/main/java/com/predic8/membrane/core/azure/api/auth/AuthenticationApi.java
```diff
@@ -49,10 +53,17 @@ public AuthenticationApi(HttpClient http, @Nullable AzureIdentity config) {
 
     public String accessToken() throws Exception {
         var response = http.call(tokenExchange()).getResponse();
-        return new ObjectMapper()
-                .readTree(response.getBodyAsStringDecoded())
-                .get("access_token")
-                .asText();
+        try {
+            return new ObjectMapper()
+                    .readTree(response.getBodyAsStringDecoded())
```
rrayst: please store the body, because the stream might be gone elsewise

core/src/main/java/com/predic8/membrane/core/azure/api/auth/AuthenticationApi.java
```diff
@@ -49,10 +53,17 @@ public AuthenticationApi(HttpClient http, @Nullable AzureIdentity config) {
...
+                    .get("access_token")
+                    .asText();
+        } catch (Exception e) {
+            log.debug(e.getMessage());
+            log.debug(tokenExchange().getRequest().toString());
```
rrayst: tokenExchange() creates a /new/ Exchange (which would not have a response set)

core/src/main/java/com/predic8/membrane/core/azure/api/auth/AuthenticationApi.java
```diff
@@ -49,10 +53,17 @@ public AuthenticationApi(HttpClient http, @Nullable AzureIdentity config) {
...
+            log.debug(tokenExchange().getResponse().toString());
```
rrayst: reading the body might not work, because `getBodyAsStringDecoded()` will already have read the body

rrayst (changes requested):

core/src/main/java/com/predic8/membrane/core/azure/api/auth/AuthenticationApi.java
```diff
@@ -52,16 +52,17 @@ public AuthenticationApi(HttpClient http, @Nullable AzureIdentity config) {
...
-                    .readTree(response.getBodyAsStringDecoded())
+                    .readTree(responseBody)
                     .get("access_token")
                     .asText();
         } catch (Exception e) {
             log.debug(e.getMessage());
-            log.debug(tokenExchange().getRequest().toString());
-            log.debug(tokenExchange().getResponse().toString());
+            log.debug(exc.getRequest().toString());
+            log.debug(exc.getResponse().getHeader().toString(), responseBody);
```
rrayst: `log.debug(string1, string2)` will not just log `string1` and `string2`, but do different interesting things

rrayst (changes requested):

rrayst (approved):

Issue #1183, opened 2024-07-11
APIProxy subpaths not matching as basepaths


t-burch:
/foo matches, but /foo/bar does not anymore.

PR #1185, opened 2024-07-12
#1183 apiproxy subpaths matching fix

predic8:


rrayst:
@predic8 why doesn't the 'test' condition go through `LanguageSupport`?

All Spel Expressions (as currently implemented in the master branch and suggested by this pull request) are always *interpreted*.

Adding these expressions to a rule key causes a slow down in *all* incoming requests.

Allowing the user to switch to Groovy would therefore increase performance significantly.

rrayst (approved):

rrayst (approved):

PR #1246, opened 2024-08-29
Implemented new Oauth2Test for JWTs

christiangoerdes:


core/src/test/java/com/predic8/membrane/core/interceptor/oauth2/OAuth2Test.java
```diff
@@ -0,0 +1,200 @@
...
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Pattern;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+
+class OAuth2Test {
```
rrayst, line 48: @christiangoerdes please also add the test to a Suite, so that it gets executed automatically

rrayst (changes requested):

rrayst (approved):

PR #1325, opened 2024-11-04
Update ServiceInvocationTest to use local rules instead of contacting BLZ Service

t-burch:


predic8 (dismissed):

core/src/test/java/com/predic8/membrane/core/IntegrationTestsWithInternet.java
```diff
@@ -32,7 +31,6 @@
         MethodTest.class,
         RegExReplaceInterceptorTest.class,
         LoadBalancingInterceptorTest.class,
```
predic8, line 33: Do you have to include the SeviceIncovationTest class?

t-burch: This was a duplicate test doing the same thing as the other I removed it

t-burch:
Actually the InterceptorInvocationTest seems to be very similar but maybe tests something different?

rrayst:
@t-burch 11 test failures...

rrayst (approved):

PR #1330, opened 2024-11-05
Updated XSLT example to use restnames instead of SQLRest

t-burch:


distribution/examples/xslt/README.md
```diff
@@ -5,110 +5,131 @@ With the XSLTInterceptor you can apply XSLT stylesheets to requests and response
...
-```
+```xml
 <router>
   <serviceProxy port="2000">
     <response>
-      <transform xslt="customer2person.xsl" />
+      <transform xslt="./reformat.xsl" />
     </response>		
-    <target host="www.thomas-bayer.com" port="80" />
+    <target url="https://api.predic8.de/restnames/name.groovy?name=Pia" />
```
predic8: Target with host an port, so that the user has to call:
localhost/restnames/name.groovy... but he can change the name.

rrayst (approved):

PR #1333, opened 2024-11-05
Enhance bootup logging to include detailed rule information

t-burch:


core/src/main/java/com/predic8/membrane/core/proxies/ProxyDisplayInfo.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    private static String proxyCustomName(Proxy proxy) {
+        if (Objects.equals(proxy.getName(), proxy.getKey().toString())) {
+            return "";
+        }
+        return "\"%s\" ".formatted(proxy.getName());
+    }
+
+    private static String proxyDisplayName(Proxy proxy) {
```
predic8, line 77: use proxy.getName();

predic8 (approved):

Issue #1349, opened 2024-11-20
api-gateway creates inconsistent redirect responses where Content-Length is 0 but still have body


precoder:
Hello,

As we develop some integration tests for our GUI, we came accross with an issue causes by inconsistent Content-Length and Body.
Here is an example body created by redirect for the login/consent endpoint:

```
HTTP/1.1 307 Temporary Redirect
Location: /auth/login/consent
Content-Type: text/html;charset=UTF-8
Content-Length: 0
Expires: Tue, 03 Jul 2001 06:00:00 GMT
Cache-Control: no-store, no-cache, must-revalidate, max-age=0
Cache-Control: post-check=0, pre-check=0
Pragma: no-cache
Connection: close

<html><head><title>Moved.</title></head><body><h1>Moved.</h1><p>This page has moved to <a href="/auth/login/consent">/auth/login/consent</a>.</p></body></html>
```

This does not cause any trouble with normal browsers however it is still an invalid combination of Content-Length and Body, therefore the Test framework do not want to accept this.

On the code I have seen this is caused by redirectToConsentPage method:
https://github.com/membrane/api-gateway/blob/master/core/src/main/java/com/predic8/membrane/core/interceptor/oauth2/processors/EmptyEndpointProcessor.java#L123

Response.redirect creates a Body as can be seen above but then "bodyEmpty()" sets the Content-Length to 0 without removing the Body.

I have seen other usages of Response.redirect where ".body("")" is called 
https://github.com/membrane/api-gateway/blob/master/core/src/main/java/com/predic8/membrane/core/interceptor/oauth2/flows/CodeFlow.java#L67

bodyEmpty() method is actually a better approach but it should also clear the Body after setting the Content-Length to 0 otherwise it creates inconsistent responses.

predic8:
@precoder thanks for the issue and the description. We will fix that.

predic8:
Fixed in master

PR #1350, opened 2024-11-20
Fix issue #1349 by setting body to empty string after 0 content-length.

precoder:
I have added the body("") into bodyEmpty method.

I can also replace other existing body("") calls on a Response.redirect builder with new bodyEmpty method in another pull request if you want to consolidate the usage.


predic8 (approved):

predic8:
Thanks @precoder for your contribution

Issue #1360, opened 2024-11-26
OpenAPI 3.1: Swagger UI does not work with split OpenAPI documents


predic8:
- See how OpenAPI is delivered
- Maybe OpenAPI can be serialized with inlined refs

predic8:
OpenAPI parser can serialize OpenAPI. But the whole model is written:

```
info: class Info {
        title: Split OpenAPI Document API
        description: null
        summary: null
        termsOfService: null
        contact: null
        license: null
        version: 1.0.0
    }
```

We do not want to have the nulls. We must find a different solution.

predic8:
For Swagger UI: maybe

getSpec(OpenAPI api) throws IOException {
        return omYaml.readTree(Json31.mapper().writeValueAsBytes(api));
}

PR #1362, opened 2024-11-26
#1360 api docs reference resolution

t-burch:


predic8 (changes requested):
We should consolidate this with the OpenAPIRecordFactory refactoring. Let's do that next week together

predic8:
Will be fixed in 6.0.0

PR #1396, opened 2024-12-09
Acme refactoring

predic8:


core/src/main/java/com/predic8/membrane/core/transport/ssl/AcmeSSLContext.java
```diff
@@ -84,7 +80,7 @@ private String[] computeHostList(String[] hostsWantedByRule, String hostsRequest
     private boolean hostMatches(String host, String certificateHost) {
         if (host.equals(certificateHost))
             return true;
-        if (certificateHost.startsWith("*.") && host.endsWith(certificateHost.substring(2)) && host.length() >= certificateHost.length() && host.codePointAt(host.length() - certificateHost.length() + 1) == 46 && isHostname(host.substring(0, host.length() - certificateHost.length() + 1)))
+        if (certificateHost.startsWith("*.") && host.endsWith(certificateHost.substring(2)) && host.length() >= certificateHost.length() && host.codePointAt(host.length() - certificateHost.length() + 1) == UTF_8_FULL_STOP && isHostname(host.substring(0, host.length() - certificateHost.length() + 1)))
```
rrayst: 46 is a constant from the TLS 1.3 spec: See the enum in https://datatracker.ietf.org/doc/html/rfc8446#section-6 . It encodes the "certificate_unknown" error.

rrayst (changes requested):

rrayst (approved):

PR #1421, opened 2024-12-17
Internal proxy example fix

t-burch:


distribution/src/test/java/com/predic8/membrane/examples/tests/InternalProxyTest.java
```diff
@@ -14,60 +14,77 @@
...
+                    .contentType(ContentType.XML)
+                    .body(regularOrder)
+                    .when()
+                    .post(BASE_URL)
+                    .then()
+                    .statusCode(200)
+                    .extract()
+                    .asString();
+
+            assertTrue(response.contains("Fallthrough switch!"));
```
predic8: Test should test for Normal or Express in response

predic8 (approved):

PR #1426, opened 2024-12-18
Cleanup logging refactoring

predic8:


core/src/main/java/com/predic8/membrane/core/interceptor/acl/Hostname.java
```diff
@@ -85,7 +85,7 @@ public boolean matches(String hostname, String ip) {
             log.debug("CanonicalHostname for {} / {} is {}", hostname, ip, canonicalHostName);
 			return Pattern.compile(schema).matcher(canonicalHostName).matches();
 		} catch (UnknownHostException e) {
-            log.warn("Could not reverse lookup canonical hostname for {} {}.", hostname, ip, e);
+            log.warn("Could not reverse lookup canonical hostname for {} {}.", hostname, ip);
```
rrayst, line 88: please also log the exception

rrayst:
In several places, logging of the exception stacktrace was removed. Was this done intentionally?

predic8:
Yes. In those places the stacktrace is of no help. The exception is catched and the reason is clear. We should keep them where it is not clear.

PR #1470, opened 2025-01-13
Refactor docs

predic8:


predic8:
Everything is already in master

PR #1474, opened 2025-01-14
Test fixes

predic8:


core/src/main/java/com/predic8/membrane/core/interceptor/lang/AbstractSetterInterceptor.java
```diff
@@ -63,24 +62,20 @@ private Outcome handleInternal(Exchange exchange, Flow flow) {
         if (!shouldSetValue(exchange, flow))
             return CONTINUE;
 
-        String msg;
         try {
             setValue(exchange, flow, exchangeExpression.evaluate(exchange, flow, Object.class));
-            return CONTINUE;
-        catch (Exception e) {
-            msg = e.getMessage();
+        } catch (Exception e) {
```
rrayst, line 66: The `Exception`'s message is swallowed, but might indicate errors in the expression (e.g. syntax errors).

rrayst (changes requested):

core/src/main/java/com/predic8/membrane/core/interceptor/lang/AbstractSetterInterceptor.java
predic8: will be fixed in the call-example branch

rrayst (dismissed):

PR #1482, opened 2025-01-16
Language improvements

predic8:


core/src/main/java/com/predic8/membrane/core/util/ConfigurationException.java
```diff
@@ -23,11 +23,25 @@
...
     public ConfigurationException(String message) {
         super(message);
     }
 
-    public ConfigurationException(String message, Exception exception) {
+    public ConfigurationException(String message, Throwable exception) {
+        super(message, exception);
+    }
+
+    public ConfigurationException(String message, String location, Throwable exception) {
```
rrayst: please remove the `location`. the information should be contained in the message.

The only place this constructor is currently used is above: `throw new ConfigurationException(ce.getMessage(), "<for in=\"%s\">".formatted(in), ce.getCause())` This effectively duplicates the `message`, because the same message is used for `ce` as well as `ce.getCause()`. It could therefore be replaced by `throw new ConfigurationException("within <for in=\"%s\">".formatted(in), ce.getCause())`.

rrayst (changes requested):

rrayst (dismissed):

rrayst (approved):

PR #1488, opened 2025-01-17
enforce interceptor.init() calling super.init()

rrayst:


core/src/test/java/com/predic8/membrane/core/acl/ACLTest.java
```diff
@@ -24,7 +24,9 @@
 
 public abstract class ACLTest extends AccessControlInterceptor {
     @Override
-    public void init() {}
+    public void init() {
```
predic8: Remove empty interceptor

predic8 (changes requested):

core/src/test/java/com/predic8/membrane/core/acl/ACLTest.java
rrayst: This is an abstract test class still in use.

predic8: Sorry. I mean: Remove empty init() :-)

predic8 (approved):

PR #1491, opened 2025-01-17
Problemdetails refactoring

predic8:


core/src/main/java/com/predic8/membrane/core/interceptor/EchoInterceptor.java
```diff
@@ -42,11 +42,11 @@ public Outcome handleRequest(Exchange exc) {
 				builder.status(204); // No Content
 			}
 		} catch (Exception e) {
-			ProblemDetails.internal(router.isProduction())
-					.component(getDisplayName())
+			log.error("Could create echo: {}",e.getMessage());
```
rrayst: "not" missing?

rrayst: additionally: This catches all `Exception`s, but only prints `e.getMessage`. - Switch to `log.error("Could not create echo: {}",e);` ?

core/src/main/java/com/predic8/membrane/core/interceptor/jwt/JwtSignInterceptor.java
```diff
@@ -76,24 +76,12 @@ private Outcome handleInternal(Exchange exc, Flow flow) {
             jws.setKeyIdHeaderValue(rsaJsonWebKey.getKeyId());
             exc.getMessage(flow).setBodyContent(jws.getCompactSerialization().getBytes());
             return CONTINUE;
-        } catch (JoseException e) {
-            log.error("Error during attempt to sign JWT payload: {}", e.getLocalizedMessage());
-            ProblemDetails.security(router.isProduction())
+        } catch (Exception e) {
```
rrayst, line 79: catching `Exception`, logging only `e.getMessage()`...

rrayst (changes requested):

rrayst (dismissed):

rrayst (dismissed):

rrayst (approved):

PR #1501, opened 2025-01-20
Move ntlm, acl examples to security directory 

t-burch:


rrayst:
Please note that the starter scripts (`*.sh`/`*.bat`) were moved to a subdirectory on a lower level. They contain `../..`/`..\..` which needs to be adjusted.

distribution/examples/security/acl/service-proxy.bat
rrayst, line 1: E.g. this file contains `..\..`

rrayst (changes requested):

distribution/examples/security/acl/service-proxy.bat
t-burch: We should wait until we replace all scripts with the new path independent versions

PR #1504, opened 2025-01-20
removed test dir

christiangoerdes:


distribution/src/test/java/com/predic8/membrane/examples/util/AssertUtils.java
```diff
@@ -0,0 +1,98 @@
+/* Copyright 2012 predic8 GmbH, www.predic8.com
```
predic8: Distribution depends on core. Do we need them here?

predic8 (changes requested):

predic8 (approved):

PR #1518, opened 2025-01-23
improved logging, ProblemDetail subTypes

rrayst:
You might want to review this PR commit-by-commit, as the changes within a single commit are quite similar.


core/src/main/java/com/predic8/membrane/core/interceptor/javascript/GraalVMJavascriptLanguageAdapter.java
```diff
@@ -32,7 +32,7 @@ public GraalVMJavascriptLanguageAdapter(Router router) {
 
     @Override
     public ProblemDetails getProblemDetails(Exception e) {
-        ProblemDetails pd = internal(router.isProduction(),"javascript");
+        ProblemDetails pd = internal(router.isProduction(),"javascript-graal");
```
predic8: Too much detail about the implementation: Caller should not know that the graal VM is in use

core/src/main/java/com/predic8/membrane/core/interceptor/javascript/RhinoJavascriptLanguageAdapter.java
```diff
@@ -31,7 +31,7 @@ public RhinoJavascriptLanguageAdapter(Router router) {
 
     @Override
     public ProblemDetails getProblemDetails(Exception e) {
-        ProblemDetails pd = ProblemDetails.internal(router.isProduction(),"javascript");
+        ProblemDetails pd = ProblemDetails.internal(router.isProduction(),"javascript-rhino");
```
predic8: ditto, do not expose rhino

core/src/main/java/com/predic8/membrane/core/interceptor/lang/AbstractSetterInterceptor.java
```diff
@@ -50,6 +50,7 @@ private Outcome handleInternal(Exchange exchange, Flow flow) {
         try {
             setValue(exchange, flow, exchangeExpression.evaluate(exchange, flow, getExpressionReturnType()));
         } catch (Exception e) {
+            log.error("While evaluating expression {} for field {}.", expression, fieldName, e);
```
predic8, line 53: Stacktrace does not provide any more details.

core/src/main/java/com/predic8/membrane/core/interceptor/server/WebServerInterceptor.java
```diff
@@ -84,7 +84,7 @@ private void normalizeDocBase() {
             try {
                 this.docBase = getAbsolutePathWithSchemePrefix(docBase);
             } catch (Exception e) {
-                log.error(e.getMessage(), e);
+                log.error("While handling docBase=", this.docBase, e);
```
predic8: {} is missing

core/src/main/java/com/predic8/membrane/core/interceptor/lang/AbstractSetterInterceptor.java
rrayst: As discussed, it provides more details, e.g. for `IOException`s. It does not provide more for XYExceptions arising from the expression execution.

`catch (Exception e)` is too broad and the potential code within try-catch too complex *not* to log the stacktrace.

predic8 (approved):

PR #1519, opened 2025-01-23
fix: make JsonGenerator not reusuable, as it is not thread safe and thereby error prone

rrayst:
…hereby error prone

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/interceptor/oauth2/BufferedJsonGenerator.java
+++ b/core/src/main/java/com/predic8/membrane/core/interceptor/oauth2/BufferedJsonGenerator.java
@@ -19,19 +19,20 @@ import com.fasterxml.jackson.core.JsonGenerator;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 
-public class BufferedJsonGenerator {
-    JsonFactory jsonFactory = new JsonFactory();
-    ByteArrayOutputStream baos = new ByteArrayOutputStream();
-    JsonGenerator jsonGenerator;
+public class BufferedJsonGenerator implements AutoCloseable{
+    private final ByteArrayOutputStream baos = new ByteArrayOutputStream();
+    private final JsonGenerator jsonGenerator;
 
     public BufferedJsonGenerator() {
         try {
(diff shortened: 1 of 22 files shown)


core/src/main/java/com/predic8/membrane/core/interceptor/oauth2/BufferedJsonGenerator.java
```diff
@@ -19,24 +19,19 @@
...
+public class BufferedJsonGenerator {
     JsonFactory jsonFactory = new JsonFactory();
     ByteArrayOutputStream baos = new ByteArrayOutputStream();
     JsonGenerator jsonGenerator;
 
-    public ReusableJsonGenerator() {
+    public BufferedJsonGenerator() {
         try {
             jsonGenerator = jsonFactory.createGenerator(baos);
         } catch (IOException e) {
```
predic8, line 30: Mark with ignore if it is not interesting

core/src/main/java/com/predic8/membrane/core/interceptor/oauth2/OAuth2Util.java
```diff
@@ -49,19 +49,18 @@ public static boolean isAbsoluteUri(String uri) {
...
         String json;
-        synchronized (jsonGen) {
-            try (JsonGenerator gen = jsonGen.resetAndGet()) {
-                gen.writeStartObject();
-                for (int i = 0; i < params.length; i += 2)
-                    gen.writeObjectField(params[i], params[i + 1]);
-                gen.writeEndObject();
-                json = jsonGen.getJson();
-            }
+        BufferedJsonGenerator jsonGen = new BufferedJsonGenerator();
```
predic8: Just an idea to make it more usable:

```
   try (BufferedJsonGenerator jsonGen = new BufferedJsonGenerator()) {
        JsonGenerator gen = jsonGen.jg()
   }
```
BufferedJsonGenerator can then close the JsonGenerator.

Maybe there is a even more ellegant way to do this.

predic8 (changes requested):

predic8 (approved):

PR #1522, opened 2025-01-24
refactor(language): Make api use ExchangeExpression

predic8:


core/src/main/java/com/predic8/membrane/core/openapi/serviceproxy/APIProxyKey.java
```diff
@@ -113,25 +103,30 @@ public String getKeyId() {
...
-            return Objects.equals(testExpr, other.testExpr);
+            return Objects.equals(exchangeExpression, other.exchangeExpression);
         }
         return false;
     }
 
     @Override
     public int hashCode() {
-        return super.hashCode() + testExpr.hashCode() + basePaths.hashCode();
+        return super.hashCode() + exchangeExpression.hashCode() + basePaths.hashCode();
```
rrayst: possible NPE. -> probably bettern to use `super.hashCode() + Objects.hashCode(exchangeExpression, basePaths)` which avoids this problem

core/src/main/java/com/predic8/membrane/core/proxies/ServiceProxyKey.java
```diff
@@ -84,49 +84,6 @@ public boolean isHostWildcard() {
         return isHostWildCard;
     }
 
-    @Override
-    public String toString() {
-        return (host.equals("*") ? "" : host + " ")
```
rrayst, line 89: Please add a note to the release notes about this (e.g. "Default naming scheme for `serviceProxy`s has changed. This might affect existing filters in log aggregation systems and/or monitoring dashboards, e.g. in Prometheus/Grafana.")

rrayst (approved):

PR #1524, opened 2025-01-25
Json2xml fix

predic8:


rrayst:
This still does not handle arrays, numbers, strings, booleans or `null`.

rrayst (approved):

PR #1529, opened 2025-01-27
refactor: log exceptions

predic8:


core/src/main/java/com/predic8/membrane/core/interceptor/HTTPClientInterceptor.java
```diff
@@ -71,6 +72,12 @@ public Outcome handleRequest(Exchange exc) {
                     .detail(msg)
                     .buildAndSetResponse(exc);
             return ABORT;
+        } catch (SocketTimeoutException e) {
+            // Details are logged further down in the HTTPClient
+            internal(router.isProduction(), getDisplayName())
+                    .detail("Target %s is not reachable.".formatted(getDestination(exc)))
```
rrayst: this prints `getDestinations().getFirst()`, which might not be the one causing the error.

rrayst (changes requested):

rrayst (approved):

PR #1558, opened 2025-01-31
feat(pathparameters):  Path parameters, soapBody Template, target method

predic8:


core/src/main/java/com/predic8/membrane/core/Router.java
```diff
@@ -357,8 +357,12 @@ private void initJmx() {
...
             throw new IllegalStateException("Hot deployment already started.");
-        if (!(beanFactory instanceof TrackingApplicationContext))
-            throw new RuntimeException("ApplicationContext is not a TrackingApplicationContext. Please set <router hotDeploy=\"false\">.");
+        if (!(beanFactory instanceof TrackingApplicationContext)) {
+            log.warn("""
+                    ApplicationContext is not a TrackingApplicationContext. Please set <router hotDeploy="false">.
+                    """);
+            return;
+        }
         if (!(beanFactory instanceof AbstractRefreshableApplicationContext))
```
rrayst, line 366: I would leave the code as is. `startHotDeployment()` does not work, if the `beanfactory` is not a `TrackingApplicationContext`.

You can call `setHotDeploy(false)` to prevent `startHotDeployment()` being called.

core/src/main/java/com/predic8/membrane/core/lang/ScriptingUtils.java
```diff
@@ -40,45 +45,62 @@ public static HashMap<String, Object> createParameterBindings(Router router, Exc
...
+
+        return params;
+    }
+
+    public static Map<String, String> extractPathParameters(Exchange exchange) {
+        if (!(exchange.getProxy() instanceof APIProxy ap))
+            return emptyMap();
+
+        try {
+            return new HashMap<>(matchTemplate(ap.getPath().getValue(), exchange.getRequestURI())); // Make lazy!
```
rrayst, line 98: what happens, if two parameters have the same key?

rrayst (approved):

core/src/main/java/com/predic8/membrane/core/Router.java
predic8: Maybe a warning is enough. That makes tests easier and membrane can work proper, even if hotdeployment is not working.

core/src/main/java/com/predic8/membrane/core/lang/ScriptingUtils.java
predic8: :-). Ugly. But how to solve that?

PR #1567, opened 2025-02-05
add copy files with maven plugin

russelmrcl:


core/pom.xml
```diff
@@ -12,499 +12,526 @@
...
-				</configuration>
-			</plugin>
+            <plugin>
+                <artifactId>maven-antrun-plugin</artifactId>
+                <version>3.0.0</version>
+                <executions>
+                    <execution>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>run</goal>
```
predic8, line 458: package

PR #1569, opened 2025-02-05
Added ApiKeyExpressionExtractor

t-burch:


core/src/test/java/com/predic8/membrane/core/interceptor/apikey/ApiKeysInterceptorTest.java
```diff
@@ -73,6 +79,17 @@ static void setup() {
...
         akiWithTwoStores.setStores(of(store, mergeStore));
+
+        akiWithExpressionExtractor = new ApiKeysInterceptor();
+        akiWithExpressionExtractor.setExtractors(of(aee));
+        akiWithExpressionExtractor.setStores(of(store));
+    }
+
+    @Test
+    void handleRequestWithKeyInsideBody() {
+        Exchange exc = new Request.Builder().body(apiKey).buildExchange();
```
predic8: get("/foo").body...

predic8 (changes requested):
see

predic8 (approved):

PR #1655, opened 2025-02-28
Type converters for scripting

predic8:


core/src/main/java/com/predic8/membrane/core/interceptor/apikey/extractors/ApiKeyExpressionExtractor.java
```diff
@@ -52,15 +53,18 @@ public String getDescription() {
         return "Expression: " + expression + ". ";
     }
 
-    @Override
```
rrayst, line 55: getter is needed for `@MCAttribute`

rrayst (approved):

rrayst (approved):

PR #1677, opened 2025-03-03
Graphql protection feature blocklist

t-burch:


core/src/main/java/com/predic8/membrane/core/graphql/GraphQLProtectionInterceptor.java
```diff
@@ -157,6 +159,15 @@ public void setMaxDepth(int maxDepth) {
         this.maxDepth = maxDepth;
     }
 
+    public FeatureBlacklist getBlacklist() {return featureBlacklist;}
```
predic8: Use allow and block not white and black.

predic8 (approved):

PR #1698, opened 2025-03-11
fix: body access for SpEL e.g. ${body}

predic8:


core/src/main/java/com/predic8/membrane/core/lang/spel/typeconverters/SpELBodyToStringTypeConverter.java
```diff
@@ -0,0 +1,35 @@
...
+
+    private static final Logger log = LoggerFactory.getLogger(SpELBodyToStringTypeConverter.class.getName());
+
+    @Override
+    public String convert(SpELBody body) {
+        try {
+            return new String(MessageUtil.getContent(body.getMessage()));
+        } catch (Exception e) {
+            log.warn("Cannot log body content", e);
+            return "";
```
rrayst: Why does this return an empty string instead of an error?

rrayst (approved):

PR #1701, opened 2025-03-12
added traffic logging

rrayst:


core/src/main/java/com/predic8/membrane/core/transport/http/ByteStreamLogging.java
```diff
@@ -43,7 +43,7 @@ public static void log(String name, byte[] b, int off, int len){
         }
         sb.append("] ");
         sb.append(new String(b,off,len));
-        log.info(sb.toString());
+        log.trace(sb.toString());
```
predic8, line 73: Is it worth to check if log is on trace level at the beginning of the method?

rrayst: This is handled via the isLoggingEnabled() method. If this returns false, the `Connecton` class won't even integrate `ByteStreamLogging` into the `Input`/`OutputStream`s.

predic8 (approved):

PR #1720, opened 2025-03-13
Refactoring of HTTP Cookie/Header related classes

bridgerdier:


core/src/main/java/com/predic8/membrane/core/http/cookie/MimeHeaders.java
```diff
@@ -39,6 +42,20 @@ public int findHeader(String string, int pos) {
 		}
 	}
 
+	public List<MessageBytes> findHeaders(String string) {
```
predic8, line 45: Unit test?

bridgerdier: According to IntelliJ, the method is already covered by plenty of tests (>600 hits)

predic8 (approved):

PR #1736, opened 2025-03-18
Intellij inspections redundancy

bridgerdier:
IntelliJ Code Inspections pointing to redundancies applied semi-automatically. Unit and Integrations tests still pass locally.


core/src/main/java/com/predic8/membrane/core/http/AbstractBody.java
```diff
@@ -233,10 +233,9 @@ public Header getTrailer() {
 	}
 
 	/**
-	 * @return true, when the body supports trailers and the trailer was therefore set.
```
predic8, line 236: Revert

core/src/main/java/com/predic8/membrane/core/http/ChunkedBody.java
```diff
@@ -301,8 +301,7 @@ public Header getTrailer() {
     }
 
     @Override
-    public boolean setTrailer(Header trailer) {
```
predic8: Revert

core/src/main/java/com/predic8/membrane/core/transport/http2/StreamInfo.java
```diff
@@ -271,9 +271,8 @@ public Header getTrailer() {
         }
 
         @Override
-        public boolean setTrailer(Header trailer) {
```
predic8: Revert

PR #1773, opened 2025-04-02
fixed JsonpathExchangeExpression: validate if body is valid json

christiangoerdes:


predic8:
@christiangoerdes fixed by #1776

PR #1803, opened 2025-04-11
Header.addAll() and BrowserMock refactoring

bridgerdier:
Added addAll operation to Header and refactored cookie handling in BrowserMock to reduce exposure to concurrency problems

## Summary by CodeRabbit

- **New Features**
  - Added support for bulk addition of header fields, enabling efficient and thread-safe updates that enhance request handling.

- **Refactor**
  - Improved header processing to achieve consistent behavior and reliable error management.
  - Streamlined cookie handling during browser interactions, boosting session reliability and overall communication stability.
  - Enhanced method documentation for better clarity on functionality and parameters.

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/http/Header.java
+++ b/core/src/main/java/com/predic8/membrane/core/http/Header.java
@@ -140,13 +140,24 @@ public class Header {
 	public Header() {
 	}
 
+	/**
+	 * Constructs a Header by reading and parsing HTTP header lines from the provided input stream.
+	 *
+	 * <p>This constructor reads the input stream line by line until an empty line is encountered,
+	 * creating and adding a HeaderField for each non-empty line. If a header line is malformed, resulting
+	 * in a StringIndexOutOfBoundsException, the error is logged and the line is skipped.</p>
+	 *
+	 * @param in the input stream containing HTTP header lines
+	 * @throws IOException if an I/O error occurs while reading the stream
+	 * @throws EndOfStreamException if the stream ends unexpectedly before a complete header is read
+	 */
 	public Header(InputStream in) throws IOException, EndOfStreamException {
 		String line;
 		while (!(line = readLine(in)).isEmpty()) {
 			try {
 				add(new HeaderField(line));
 			} catch (StringIndexOutOfBoundsException sie) {
-                log.error("Header read line that caused problems: {}", line);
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/predic8/membrane/core/http/Header.java
```diff
@@ -170,10 +181,31 @@ public void add(String key, String val) {
...
 		fields.add(field);
 	}
 
+	/**
+	 * Adds all header fields from the given list to this header in a thread-safe manner.
+	 *
+	 * @param fields the list of header fields to add
+	 */
+	public void addAll(List<HeaderField> fields) {
+		synchronized (this.fields) {
```
rrayst: The `Header` class is not thread-safe. Therefore, `synchronized` can be removed.

rrayst (changes requested):

core/src/main/java/com/predic8/membrane/core/http/Header.java
bridgerdier: Resolved in [8f346d7](https://github.com/membrane/api-gateway/pull/1803/commits/8f346d7f55fb7c98e039207cb8b4149ec9073706)

core/src/test/java/com/predic8/membrane/core/interceptor/oauth2/client/BrowserMock.java
```diff
@@ -168,9 +168,12 @@ public class BrowserMock implements Function<Exchange, Exchange> {
     private static void addCookiesToExchange(Exchange exc, Map<String, String> cookies) {
         if (cookies == null)
             return;
-        exc.getRequest().getHeader().addAll(cookies.entrySet().stream()
-                .map(BrowserMock::createCookieHeaderField)
-                .toList());
+        Header header = exc.getRequest().getHeader();
+        synchronized (header) {
```
rrayst: TL/DR: Please remove the `synchronized(header)`.

The header here is also used only by one thread. (Accessing an `Exchange` or one of its sub-objects from multiple threads in parallel is always not safe. And this is not done here: Each thread has one `Exchange` for itself.)

If something needed protection, it was `browserMock.cookie`: This is actually read/written to by multiple threads.

rrayst (changes requested):

rrayst (approved):

PR #1862, opened 2025-05-08
Guava map

russelmrcl:
## Summary by CodeRabbit

- **New Features**
  - Added the ability to configure the expiration time (in seconds) for idempotency keys, with a default of 3600 seconds.

- **Improvements**
  - Enhanced idempotency key handling to automatically remove old keys, improving memory usage and reliability.
  - Updated documentation to provide a clearer example for key expressions.
  - Changed HTTP response code for duplicate keys from 400 to 409.

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/interceptor/idempotency/IdempotencyInterceptor.java
+++ b/core/src/main/java/com/predic8/membrane/core/interceptor/idempotency/IdempotencyInterceptor.java
@@ -14,6 +14,8 @@
 
 package com.predic8.membrane.core.interceptor.idempotency;
 
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
 import com.predic8.membrane.annot.MCAttribute;
 import com.predic8.membrane.annot.MCElement;
 import com.predic8.membrane.annot.Required;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/predic8/membrane/core/interceptor/idempotency/IdempotencyInterceptor.java
```diff
@@ -111,11 +115,32 @@ public void setLanguage(Language language) {
         this.language = language;
     }
 
+    /**
+     * @description Time in seconds after which idempotency keys automatically expire.
+     * Useful to avoid memory leaks in long-running systems.
+     * Common values:
+     * <ul>
```
predic8, line 122: Default 3600 
setExpiration

core/src/main/java/com/predic8/membrane/core/interceptor/idempotency/IdempotencyInterceptor.java
```diff
@@ -47,12 +48,17 @@ public class IdempotencyInterceptor extends AbstractInterceptor {
     private String key;
     private ExchangeExpression exchangeExpression;
     private Language language = SPEL;
-    private final Map<String, Boolean> processedKeys = new ConcurrentHashMap<>();
+    private int seconds = 600;
```
predic8: 409

predic8 (changes requested):

PR #1863, opened 2025-05-09
Implement expression support in Target

t-burch:


core/src/main/java/com/predic8/membrane/core/proxies/AbstractServiceProxy.java
```diff
@@ -183,6 +206,24 @@ public String getMethod() {
...
+        public ExchangeExpression getExchangeExpression() {
+            return exchangeExpression;
+        }
+
+        public ExchangeExpression.Language getLanguage() {
+            return language;
+        }
+
+        /**
+         * @description the language of the 'test' condition
```
predic8: test cond.

PR #1867, opened 2025-05-16
Api gateway cuncurrency patch

Lucamadio:
Replaced generic classes with corrensponding of concurrency package (like ConcurrentLinkedQueue, AtomicInteger...) and optimized code using lambdas.
Using the concurrency classes allowed us to remove several synchronized blocks, and by profiling the application with VisualVM, we observed an improvement in resource usage and release

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/exchangestore/LimitedMemoryExchangeStore.java
+++ b/core/src/main/java/com/predic8/membrane/core/exchangestore/LimitedMemoryExchangeStore.java
@@ -14,18 +14,35 @@
 
 package com.predic8.membrane.core.exchangestore;
 
-import com.predic8.membrane.annot.*;
-import com.predic8.membrane.core.exchange.*;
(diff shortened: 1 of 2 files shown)


predic8:
@Lucamadio thanks for this promising PR. Your PR will be review by coworker as soon as he is back from vacation.
Looking forward to see improvements.

rrayst (approved):

rrayst (commented):
progress update: I'm still on this. (Concurrency is hard...) I still need to think about `HttpEndpointListener`.

(Note: I think `limitedMemoryExchangeStore.lastModification` should probably made `volatile`, as it is accessed (read-) without `synchronized` now.)

Lucamadio:
Thanks for the kind progress update. 
Regarding the <code>lastModification</code> variable, in the class it is always used within synchronized blocks, whereas its getter method 
(from which we removed the synchronized keyword), I couldn’t find any reference where it is actually used. 
Maybe the method could even be removed.

PR #1871, opened 2025-05-21
Improve error message on misplacement of the openapi element

t-burch:


core/src/main/java/com/predic8/membrane/core/cli/RouterCLI.java
```diff
@@ -107,7 +107,12 @@ public static String getExceptionMessageWithCauses(Throwable throwable) {
                 default -> initRouterByConfig(commandLine);
             };
         } catch (InvalidConfigurationException e) {
-            log.error("Fatal error: {}", concatMessageAndCauseMessages(e));
+            String errorMsg = concatMessageAndCauseMessages(e);
+            if (errorMsg.contains("Invalid content was found starting with element '{\"http://membrane-soa.org/proxies/1/\":openapi}'.")) {
```
predic8: Make shorter to match even when the message is slightly different. If the jdk language is german the if does not work

core/src/main/java/com/predic8/membrane/core/cli/RouterCLI.java
```diff
@@ -116,6 +121,10 @@ public static String getExceptionMessageWithCauses(Throwable throwable) {
         return null;
     }
 
+    private static boolean isOpenAPIMisplacedError(String errorMsg) {
```
predic8: Test

PR #1885, opened 2025-06-03
REST2SOAPInterceptor Response XSLT Empty

Lucamadio:
On response Flow if mapping.responseXSLT is empty  return directly CONTINUE. There is not way to avoid an XSL trasformation on response flow for an exchange which the mapping regex has been correctly evaluated on request flow.

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/interceptor/rest/REST2SOAPInterceptor.java
+++ b/core/src/main/java/com/predic8/membrane/core/interceptor/rest/REST2SOAPInterceptor.java
@@ -196,7 +196,7 @@ public class REST2SOAPInterceptor extends SOAPRESTHelper {
     public Outcome handleResponseInternal(Exchange exc) throws Exception {
         Mapping mapping = getRESTURL(exc);
         log.debug("restURL: {}",mapping);
-        if (getRESTURL(exc) == null)
+       if (mapping == null || mapping.getResponseXSLT().isEmpty())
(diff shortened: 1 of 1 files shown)


predic8:
Hi @Lucamadio,
thanks for the PR. Just to let you know in Membrane 6.1.0 it will be much simpler to do REST2SOAP Mapping like in the sample below. 

```
		<api port="2000" method="GET">
			<path>/cities/{city}</path>
			<request>
				<soapBody version="1.1"><![CDATA[
                    <getCity xmlns="https://predic8.de/cities">
                        <name>${pathParam.city}</name>
                    </getCity>
                ]]></soapBody>
			</request>
			<response>
				<template contentType="application/json">
					{
						"country": "${property.country}",
						"population": "${property.population}"
					}
				</template>
				<setProperty name="country" value="${//country}" language="xpath"/>
				<setProperty name="population" value="${//population}" language="xpath"/>
			</response>
			<target method="POST" url="https://www.predic8.de/city-service"/>
		</api>
```

predic8 (approved):

Lucamadio:
Thanks @predic8.
Rest2Soap in version 6.1.0 seems to be highly customizable, i will try as soon as possible. My PR is due to the fact that I'm using rest2soap, which includes asynchronous endpoints with empty bodies responses, and I haven't found any other way to avoid the transformation in the response flow except by setting an empty responseXSL. 
Thanks again.

PR #1899, opened 2025-06-12
Add SetCookiesInterceptor

christiangoerdes:
- Implement SetCookiesInterceptor with support for multiple cookie definitions
- Define CookieDef class with attributes: name, value, domain, path, maxAge, expires, secure, httpOnly, and SameSite
- Support building standard Set-Cookie header strings with all attributes
- Add unit tests for header building with default and full attribute sets, including SameSite variations

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/interceptor/lang/SetCookiesInterceptor.java
+++ b/core/src/main/java/com/predic8/membrane/core/interceptor/lang/SetCookiesInterceptor.java
@@ -1,0 +1,165 @@
+package com.predic8.membrane.core.interceptor.lang;
+
+import com.predic8.membrane.annot.MCAttribute;
+import com.predic8.membrane.annot.MCChildElement;
+import com.predic8.membrane.annot.MCElement;
+import com.predic8.membrane.core.exchange.Exchange;
+import com.predic8.membrane.core.interceptor.AbstractInterceptor;
+import com.predic8.membrane.core.interceptor.Outcome;
+
+import java.util.ArrayList;
+import java.util.List;
+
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/predic8/membrane/core/interceptor/lang/SetCookiesInterceptor.java
```diff
@@ -0,0 +1,177 @@
...
+import com.predic8.membrane.core.interceptor.AbstractInterceptor;
+import com.predic8.membrane.core.interceptor.Outcome;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static com.predic8.membrane.core.interceptor.Outcome.CONTINUE;
+
+/**
+ * Interceptor that adds one or more Set-Cookie headers on the response.
```
predic8: not interceptor

core/src/main/java/com/predic8/membrane/core/interceptor/lang/SetCookiesInterceptor.java
```diff
@@ -0,0 +1,224 @@
...
+import com.predic8.membrane.core.interceptor.Outcome;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static com.predic8.membrane.core.interceptor.Outcome.CONTINUE;
+
+
+/**
+ * @description Plugin that adds one or more Set-Cookie headers on the HTTP response.
```
predic8: Advantage compared to doing it by hand? Computing expire. Setting, Building cookie string.

predic8 (approved):

PR #1946, opened 2025-07-02
Fixes #1904.

### LINKED ISSUE #1904: JSONProtection: Add Prototype Pollution
Do research

Detect  "__proto__" fields

Log as warning

Discuss: remove or block

rrayst:


predic8:
Think that blocking is better than filtering it out as __proto should not be in the JSON.

Maybe change option to: blockProto

predic8 (approved):

PR #1974, opened 2025-07-10
fix: OPTIONS request with no body can cause the thread to hang

predic8:
## Summary by CodeRabbit

* **Refactor**
  * Improved internal structure and readability of the HTTP server handler, resulting in clearer and more maintainable request processing without changing user-facing behavior.
  * Enhanced consistency in HTTP method handling by using predefined constants.
  * Updated HTTP request body detection logic for more accurate processing.

* **Bug Fixes**
  * Enhanced handling of HTTP OPTIONS requests to more accurately determine when a request body should be present, especially when using Content-Length or Transfer-Encoding headers.

* **Tests**
  * Added new tests to verify correct body handling for HTTP OPTIONS requests under various header conditions.

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/transport/http/HttpServerHandler.java
+++ b/core/src/main/java/com/predic8/membrane/core/transport/http/HttpServerHandler.java
@@ -19,6 +19,7 @@ import com.predic8.membrane.core.http.*;
 import com.predic8.membrane.core.transport.http2.*;
 import com.predic8.membrane.core.transport.ssl.*;
 import com.predic8.membrane.core.util.*;
+import org.jetbrains.annotations.*;
 import org.slf4j.*;
 
 import javax.net.ssl.*;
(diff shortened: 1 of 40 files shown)


predic8:
@coderabbitai is there danger that this PR breaks something?

core/src/main/java/com/predic8/membrane/core/transport/http/HttpServerHandler.java
```diff
@@ -27,338 +28,396 @@
...
+            return terminateWithConnection(con);
+        }
+
+        return processHttp1Request(con);
+    }
+
+    private boolean waitForIncomingData() throws IOException {
+        endpointListener.setIdleStatus(sourceSocket, true);
+        boolean endOfStream = isEndOfStream();
+        endpointListener.setIdleStatus(sourceSocket, false);
```
rrayst: This leaks, as expected, the socket in case of a non-IOException within `isEndOfStream()`.

The contract is: If you call `eL.setIdleStatus(someSocket, true);` you *have to* ensure `eL.setIdleStatus(someSocket, false);` is called.

core/src/main/java/com/predic8/membrane/core/transport/http/HttpServerHandler.java
```diff
@@ -27,338 +28,396 @@
...
+        // Prepare
+        srcReq = new Request();
+        if (con != null) {
+            exchange.setTargetConnection(con);
+            con = null;
+        }
+
+        readAndParseRequest();
+        process();
+        return determineConnectionContinuation(con);
```
rrayst, line 214: `con` is always null.

core/src/main/java/com/predic8/membrane/core/transport/http/HttpServerHandler.java
```diff
@@ -27,338 +28,396 @@
...
+            srcReq.getHeader().add(CONNECTION,
+                    srcReq.getHeader().getProxyConnection());
+            srcReq.getHeader().removeFields(PROXY_CONNECTION);
+        }
+    }
+
+    private @NotNull RequestProcessingResult determineConnectionContinuation(Connection con) {
+        if (srcReq.isCONNECTRequest()) {
+            log.debug("stopping HTTP Server Thread after establishing an HTTP connect");
+            return new RequestProcessingResult(true, con);
```
rrayst: `new RequestProcessingResult(true, ` -> `terminateWithConnection(`

core/src/main/java/com/predic8/membrane/core/transport/http/HttpServerHandler.java
predic8: @coderabbitai is that really the case?

core/src/main/java/com/predic8/membrane/core/transport/http/HttpServerHandler.java
```diff
@@ -188,24 +187,21 @@ private RequestProcessingResult processSingleRequest(Connection con) throws Exce
...
-            if (srcIn.read() == -1)
+            if (srcIn.read() != -1) {
+                srcIn.reset();
                 return true;
-            srcIn.reset();
+            }
             return false;
         } catch (IOException e) {
-            return true;
+            return false;
```
rrayst: throw `IOException`?

PR #2033, opened 2025-08-01
refactor code

russelmrcl:


core/src/main/java/com/predic8/membrane/balancer/client/LBNotificationClient.java
```diff
@@ -81,7 +78,7 @@ public void run(String[] args) throws Exception {
 	private Response notifiyClusterManager() throws Exception {
 		try (HttpClient client = new HttpClient()) {
 			Exchange exc = new Exchange(null);
-			Request r = MessageUtil.getPostRequest(getRequestURL());
+			Request r = new Request.Builder().post(getRequestURL()).build();
 			r.setBodyContent(new byte[0]);
```
predic8: use static import: r = post(...

predic8 (approved):

PR #2094, opened 2025-08-26
fix: absolute paths in request URI

predic8:
## Summary by CodeRabbit

- New Features
  - Centralized URI resolution that preserves path and query; CONNECT requests retain the full URI.
  - New public accessors: raw fragment and path-with-query (fragment is not propagated to backends).
  - Ability to record an original request URI for rewrite/diagnostics.

- Bug Fixes
  - More reliable construction of forwarded and absolute-request URLs and improved invalid-URI handling.

- Tests
  - Reorganized tests, reduced visibility, and added coverage for path/query handling and absolute-URL scenarios.

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/interceptor/DispatchingInterceptor.java
+++ b/core/src/main/java/com/predic8/membrane/core/interceptor/DispatchingInterceptor.java
@@ -55,7 +55,7 @@ public class DispatchingInterceptor extends AbstractInterceptor {
             try {
                 exc.getDestinations().add(getForwardingDestination(exc));
             } catch (Exception e) {
-                internal(router.isProduction(),getDisplayName())
+                internal(router.isProduction(), getDisplayName())
                         .detail("Could not get forwarding destination to dispatch request")
                         .exception(e)
                         .buildAndSetResponse(exc);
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/predic8/membrane/core/interceptor/DispatchingInterceptor.java
```diff
@@ -96,7 +96,7 @@ protected String getAddressFromTargetElement(Exchange exc) throws MalformedURLEx
             return targetURL;
         }
         if (p.getTargetHost() != null) {
-            return new URL(p.getTargetScheme(), p.getTargetHost(), p.getTargetPort(), exc.getRequest().getUri()).toString();
+            return new URL(p.getTargetScheme(), p.getTargetHost(), p.getTargetPort(), exc.getRequestURI()).toString();
```
rrayst: why was this changed? `exc.getRequestURI()` has interesting semantics...

core/src/main/java/com/predic8/membrane/core/interceptor/DispatchingInterceptor.java
```diff
@@ -43,6 +45,8 @@ public class DispatchingInterceptor extends AbstractInterceptor {
 
     private static final Logger log = LoggerFactory.getLogger(DispatchingInterceptor.class.getName());
 
+    private final URIFactory uriFactory = new URIFactory(true);
```
rrayst: -> `router.getUriFactory()`

rrayst (dismissed):

rrayst (dismissed):

PR #2110, opened 2025-09-01
Problem details log key

russelmrcl:
## Summary by CodeRabbit

- New Features
  - Added a configurable option to include a log reference key in error responses.
  - In production, error responses no longer include a log reference key by default; it can be explicitly enabled via configuration.


rrayst:
* call to `logKey(...)`
* when `logKey==false`, disabling logging a logkey ("if production" -> "if production && logKey")

PR #2118, opened 2025-09-03
README.md YAML URL Template

predic8:
## Summary by CodeRabbit

* Documentation
  * Added guidance for path rewriting using URI templates (e.g., mapping /fruit/{id} via path parameters).
  * Introduced a beta YAML-based API configuration example with key sections and a reference link.
  * Improved YAML example formatting and expanded run instructions, including opening API docs and the sample UI.
  * Added a sample YAML demonstrating a logging API with a log interceptor and external target for local testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,12 +16,37 @@ Solve even complex custom API requirements with simple configurations.
 </api>
 ```
 
+**Path Rewriting with an URI Template:**
+```xml
+<api port="2000">
+    <path>/fruit/{id}</path>
+    <target url="https://api.predic8.de/shop/v2/products/${pathParam.id}"/>
+</api>
+```
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -16,13 +16,38 @@ Solve even complex custom API requirements with simple configurations.
...
+metadata:
+  name: log
+spec:
+  port: 2000
+  interceptors:
+    - log:
+        message: Header ${header}
+  target:
+    url: https://api.predic8.de
+```
```
rrayst, line 47: is the YAML config intentionally semantically different from the XML snippet above?

rrayst (changes requested):

README.md
predic8: Yes

rrayst (approved):

PR #2123, opened 2025-09-06
Dispatching: Error Details for Illegal Path

predic8:
## Summary by CodeRabbit

- Bug Fixes
  - Invalid request paths now return 400 Bad Request instead of 500.
  - Error responses consistently use standardized Problem Details fields (title, type, detail, instance) and include extra context (path and index) for clearer troubleshooting.

- Tests
  - Added tests verifying JSON Problem Details for invalid URIs and updated expectations to reflect 400 responses.

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/interceptor/DispatchingInterceptor.java
+++ b/core/src/main/java/com/predic8/membrane/core/interceptor/DispatchingInterceptor.java
@@ -14,9 +14,11 @@
 package com.predic8.membrane.core.interceptor;
 
 import com.predic8.membrane.annot.*;
+import com.predic8.membrane.core.exceptions.*;
 import com.predic8.membrane.core.exchange.*;
 import com.predic8.membrane.core.openapi.util.*;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/predic8/membrane/core/interceptor/DispatchingInterceptor.java
```diff
@@ -68,6 +79,14 @@ public Outcome handleRequest(Exchange exc) {
         return CONTINUE;
     }
 
+    private static @NotNull String getMessageForURISyntaxException(Exchange exc, URISyntaxException e) {
+        var uri = exc.getRequestURI();
+        if (e.getIndex() >= 0 && e.getIndex() < uri.length()) {
+            return "The request path contains an invalid character '%s' at pos %d".formatted(uri.charAt(e.getIndex()), e.getIndex());
```
rrayst, line 85: maybe add a hint to use `<uriFactory allowIllegalCharacters="true">`?

rrayst (approved):

PR #2172, opened 2025-09-24
update interceptor documentation

russelmrcl:
## Summary by CodeRabbit

* **Documentation**
  * Clarified usage scope of the load balancing interceptor and fixed capitalization/terminology.
  * Improved conditional interceptor docs: clarified behavior, added a SpEL example, updated default language and phrasing for test conditions.
  * Polished return interceptor docs: tightened wording for status code and content type and emphasized key parameters.
  * No functional changes or public API modifications.


core/src/main/java/com/predic8/membrane/core/interceptor/flow/IfInterceptor.java
```diff
@@ -29,13 +29,12 @@
 
 /**
  * @description <p>
- * The "if" interceptor supports conditional execution of nested plugins.
+ * The if allows conditional execution of nested interceptors.
```
predic8: If

PR #2253, opened 2025-11-01
XML Namespace Declarations for XPath Expressions

predic8:
## Summary by CodeRabbit

* **New Features**
  * XML namespace support for dynamic expression evaluation across interceptors
  * Request builder: xml(body) for creating XML requests
  * XPath evaluation now honors configured XML namespaces

* **Documentation**
  * New XML Namespaces example with runnable samples and scripts

* **Chores**
  * Hardened, XXE‑safe XML parser added for safer parsing
  * Expression framework updated to propagate XML configuration

* **Tests**
  * New tests covering namespace handling and XPath/JSONPath expression behavior

### DIFF
--- a/distribution/examples/xml/namespaces/proxies.xml
+++ b/distribution/examples/xml/namespaces/proxies.xml
@@ -4,11 +4,13 @@
               xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-3.0.xsd
 					    http://membrane-soa.org/proxies/1/ http://membrane-soa.org/schemas/proxies-1.xsd">
 
-    <!-- Declare the used namespaces here -->
-    <namespaces id="ns">
-        <namespace prefix="per" uri="https://predic8.de/person"/>
-        <namespace prefix="ns1" uri="https://predic8.de/address"/>
-    </namespaces>
+    <!-- Reference with <spring:ref bean="ns"/> -->
+    <xmlConfig id="ns">
+        <namespaces>
+            <namespace prefix="per" uri="https://predic8.de/person"/>
+            <namespace prefix="ns1" uri="https://predic8.de/address"/>
+        </namespaces>
+    </xmlConfig>
 
     <router>
 
@@ -19,12 +21,18 @@
         </api>
 
         <api port="2000">
-            <rateLimiter language="xpath" requestLimit="10" keyExpression="//per:person/@id">
-                <spring:ref bean="ns"/>
+            <rateLimiter language="xpath" requestLimit="10" keyExpression="withNs(ns,//per:person/@id)">
+                <!-- Local declaration of namespaces -->
+                <xmlConfig>
+                    <namespaces id="ns">
+                        <namespace prefix="per" uri="https://predic8.de/person"/>
+                        <namespace prefix="ns1" uri="https://predic8.de/address"/>
+                    </namespaces>
+                </xmlConfig>
             </rateLimiter>
             <request>
                 <log message="Id: ${/per:person/@id}" language="xpath">
-                    <spring:ref bean="ns"/>
+                    <spring:ref bean="ns"/> <!-- References the xmlConfig declaration at the top of this file -->
                 </log>
                 <setProperty name="name" value="${/per:person/per:name}" language="xpath">
                     <spring:ref bean="ns"/>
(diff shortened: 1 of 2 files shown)


distribution/src/test/java/com/predic8/membrane/examples/withoutinternet/test/xml/namespaces/NamespacesExampleTest.java
```diff
@@ -0,0 +1,28 @@
...
+
+public class NamespacesExampleTest extends AbstractSampleMembraneStartStopTestcase {
+
+    @Override
+    protected String getExampleDirName() {
+        return "xml/namespaces";
+    }
+
+    // TODO Implement
+
```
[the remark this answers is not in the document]
predic8: generate test implementation

distribution/examples/xml/namespaces/membrane.cmd
```diff
@@ -0,0 +1,17 @@
+@echo off
+setlocal
+set "dir=%CD%"
```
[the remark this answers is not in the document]
predic8: @coderabbitai can you create an issue for that

core/src/main/java/com/predic8/membrane/core/lang/ExchangeExpression.java
```diff
@@ -48,12 +49,46 @@ enum Language {GROOVY, SPEL, XPATH, JSONPATH}
...
+     */
+    static ExchangeExpression expression(Interceptor interceptor, Language language, String expression) {
         return switch (language) {
-            case GROOVY -> new GroovyExchangeExpression(router, expression);
+            case GROOVY -> new GroovyExchangeExpression(interceptor, expression);
             case SPEL -> new SpELExchangeExpression(expression,null);
-            case XPATH -> new XPathExchangeExpression(expression);
+            case XPATH -> new XPathExchangeExpression(interceptor,expression);
             case JSONPATH -> new JsonpathExchangeExpression(expression);
         };
```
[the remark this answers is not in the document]
predic8: @coderabbitai There is no trace in the git history that something other than null was handed previously. Can you give me a pointer or explain?

core/src/main/java/com/predic8/membrane/core/interceptor/lang/Namespaces.java
```diff
@@ -0,0 +1,102 @@
...
+package com.predic8.membrane.core.interceptor.lang;
+
+import com.predic8.membrane.annot.*;
+
+import javax.xml.namespace.*;
+import java.util.*;
+
+import static javax.xml.XMLConstants.NULL_NS_URI;
+
+@MCElement(name="namespaces", topLevel = true)
```
rrayst: maybe change to `name="xmlNamespaces"`, since this is XML-specific?

core/src/main/java/com/predic8/membrane/core/lang/ExchangeExpression.java
```diff
@@ -48,12 +49,47 @@ enum Language {GROOVY, SPEL, XPATH, JSONPATH}
...
-    static ExchangeExpression expression(Router router, Language language, String expression) {
+    /**
+     * Clients of this class should pass an interceptor if possible. Otherwise use the InterceptorAdapter to wrap it.
+     * There is no convenience method on purpose to make the clients pass the interceptor. From the interceptor you can always get the router.
+     * @param interceptor
+     * @param language
+     * @param expression
+     * @return
+     */
+    static ExchangeExpression expression(Interceptor interceptor, Language language, String expression) {
```
rrayst, line 61: why do we need the interceptor here? is this the new god-object?

core/src/main/java/com/predic8/membrane/core/util/ExceptionUtil.java
```diff
@@ -25,4 +26,14 @@ public static String concatMessageAndCauseMessages(Throwable throwable) {
         } while (throwable != null);
         return sb.toString();
     }
+
+    public static Throwable getRootCause(Throwable t) {
```
rrayst, line 30: why do we throw the outer exceptions away? usually, they are there to provide context.

core/src/main/java/com/predic8/membrane/core/lang/ExchangeExpression.java
rrayst: maybe we should declare namespaces globally once instead of adjusting every element which potentially contains an expression.

maybe we could also alternatively do something like `${withNamespaces(globalNsBean, //per:person/per:name)}`

core/src/main/java/com/predic8/membrane/core/util/ExceptionUtil.java
predic8: Generally you are right. In this case the innermost exception has the message that describes the problem very precise. Let's discuss this in the kitchen.

PR #2261, opened 2025-11-03
Move AccessLogInterceptorService to log.access package and add Test to ensure single Logger call

christiangoerdes:
## Summary by CodeRabbit

* **Refactor**
  * Access logging code reorganized into a narrower package namespace.

* **Tests**
  * Added an architecture test to enforce a single logger call in the access logging package.

* **Chores**
  * Example logging configuration updated to reference the reorganized access logging namespace.


rrayst:
please also adjust `distribution/examples/logging/access/log4j2_access.xml`

predic8 (approved):

PR #2281, opened 2025-11-07
Add <api> name genaration from OpenApi title

christiangoerdes:
## Summary by CodeRabbit

* **Bug Fixes**
  * OpenAPI services without explicit names now receive human-friendly names derived from the API title.
  * If an x-api-id extension is present it is used as the service name.
  * Names are sanitized to remove non-printable characters and truncated to 32 characters.
  * When multiple APIs are present, the name appends " +N more".
  * If no title is available, the name defaults to "Unnamed OpenAPI".


core/src/main/java/com/predic8/membrane/core/openapi/serviceproxy/APIProxy.java
```diff
@@ -112,6 +112,26 @@ private void initOpenAPI() {
         if (getFirstInterceptorOfType(OpenAPIPublisherInterceptor.class).isEmpty()) {
             interceptors.addFirst(new OpenAPIPublisherInterceptor(apiRecords));
         }
+
+        assignOpenAPIName();
+    }
+
+    void assignOpenAPIName() {
+        if (name.isEmpty() && !apiRecords.isEmpty()) {
```
predic8: guard

core/src/main/java/com/predic8/membrane/core/openapi/serviceproxy/APIProxy.java
```diff
@@ -112,6 +112,26 @@ private void initOpenAPI() {
...
+            var info = apiRecords.values().iterator().next().getApi().getInfo();
+            String title = info != null ? info.getTitle() : null;
+            if (title != null && !title.isBlank()) {
+                name = title.replaceAll("[^a-zA-Z0-9-_]", "").trim();
```
predic8: there is a method in StringUtils or another util that does that

predic8 (changes requested):

core/src/main/java/com/predic8/membrane/core/openapi/serviceproxy/APIProxy.java
predic8: Maybe we should try to use the title as it is as a name. Maybe that's better

core/src/main/java/com/predic8/membrane/core/openapi/serviceproxy/APIProxy.java
```diff
@@ -112,6 +113,25 @@ private void initOpenAPI() {
         if (getFirstInterceptorOfType(OpenAPIPublisherInterceptor.class).isEmpty()) {
             interceptors.addFirst(new OpenAPIPublisherInterceptor(apiRecords));
         }
+
+        assignOpenAPIName();
+    }
+
+    void assignOpenAPIName() {
```
predic8: Rewrite to return a String without a side effect

core/src/main/java/com/predic8/membrane/core/openapi/serviceproxy/APIProxy.java
```diff
@@ -112,6 +113,25 @@ private void initOpenAPI() {
...
+        if(!name.isEmpty() || apiRecords.isEmpty()) return;
+
+        var info = apiRecords.values().iterator().next().getApi().getInfo();
```
predic8, line 122: If there is a x-api-id in the openapi use that:

See:
https://opensource.zalando.com/restful-api-guidelines/#215

PR #2357, opened 2025-12-02
Support rewriter on internal destinations

t-burch:
## Summary by CodeRabbit

* **Bug Fixes**
  * Internal URLs are now handled like HTTP URLs during forwarding, improving consistency while preserving fallback behavior.

* **Refactor**
  * Path resolution when base path is empty/root now uses URI-based resolution for consistent normalization.

* **Tests**
  * Internal routing tests updated to use the explicit internal://... URL format.

* **Documentation**
  * Roadmap updated: internal:<name> removed (only internal://<name> supported) and noted interface changes including a client call signature and removal of WADLInterceptor.


core/src/test/resources/internal-invocation/proxies.xml
```diff
@@ -74,13 +74,13 @@
 		<serviceProxy name="FullChain0" port="3028">
 			<interceptor refid="Mock1" />
 			<interceptor refid="Mock2" />
-			<target url="internal:FullChain1" />
+			<target url="internal://FullChain1" />
```
predic8, line 77: Excellent work!

We brake the compatibility here. internal:foo is not longer working. I think that's fine as long as we document that for the 7.0.0. Can you add a ready formulated text into the ROADMAP.md so we can copy that later for the release notes under "Breaking changes?"

t-burch: Yes I'll add it. Though it will only apply to people that have been using internal for a while as the membrane-api.io documentation uses the new syntax already

rrayst (approved):

PR #2373, opened 2025-12-05
Support reading and concatenating multiple header values in SpE…

predic8:
…L expressions

## Summary by CodeRabbit

* **New Features**
  * Headers now return all matching values and present them as a single comma-separated string for easier consumption.

* **Tests**
  * Added test coverage validating multi-valued header retrieval and aggregation behavior.


rrayst:
1 failing test.

core/src/main/java/com/predic8/membrane/core/lang/spel/spelable/SpELHeader.java
```diff
@@ -29,15 +31,21 @@ public SpELHeader(Header header) {
 
     @Override
     public TypedValue read(EvaluationContext context, Object target, String name) {
-        var v = header.getFirstValue(name);
-        if (v != null)
-            return new TypedValue(v);
-        v = header.getFirstValue(camelToKebab(name));
+        List<HeaderField> values = header.getValues(name);
+        if (values.isEmpty()) {
+            values = header.getValues(camelToKebab(name));
```
rrayst, line 36: Can we make accessing the headers via a camel name consistent? If `camelToKebab()` should be called, it should be *always* called.

rrayst: Does this work with removal? How do I implement "if header xApiKey is present, remove xApiKey ."?

predic8: For me the camelCase access is also wired. I just took the existing logik. Maybe we should discuss removing it. I do not know where we access header like header.contentType ?

@coderabbitai what do you think?

rrayst: @coderabbitai isn't this a potential security issue misleading the API gateway, if the client sends "X-Api-Key: abc\nxApiKey: xyz\n" as HTTP headers?

predic8: @coderabbitai With option 1 a user tries a real standardized HTTP header name and it fails. 

The security problem is more with confusing the content type and avoiding validation or missleading jsonpath expressions.

+1 for removing that camelCase feature. I think it is not used in any example.

predic8:
All of that is now implemented in #2386

PR #2442, opened 2025-12-12
Fixes #2439

### LINKED ISSUE #2439: YAML Samples: Version numer is not automatically updated
in 7.0.1 there is still:

# yaml-language-server: $schema=https://www.membrane-api.io/v7.0.0.json

russelmrcl:
## Summary by CodeRabbit

* **Chores**
  * Adjusted the release build workflow so the schema URL update runs later in the job (after checkout), preserving existing behavior while improving step ordering during release builds.

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -61,7 +61,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           ref: ${{ github.ref }}
-      - name: Re-apply $schema URL update for artifacts
+      - name: Update $schema URL in YAML files
         run: |
           VERSION='${{ steps.get_version.outputs.VERSION }}'
           echo "Updating schema URLs in YAML files to v${VERSION}.json"


.github/workflows/release-build.yml
```diff
@@ -72,6 +71,15 @@ jobs:
       - uses: actions/checkout@v3
         with:
           ref: ${{ github.ref }}
+
+      - name: Re-apply $schema URL update for artifacts
+        run: |
+          VERSION='${{ steps.get_version.outputs.VERSION }}'
+          echo "Re-applying schema URLs in YAML files to v${VERSION}.json"
+          find . \( -name "apis.yaml" -o -name "apis.yml" \) -print0 | while IFS= read -r -d '' file; do
+            sed -i "s|\$schema=https://www\.membrane-api\.io/v[0-9\.]\+\.json|\$schema=https://www.membrane-api.io/v${VERSION}.json|g" "$file"
```
rrayst, line 69: @predic8 should we enable non-numeric version identifiers going forward?

predic8: For now we can live with it. I also depends on the effort.

PR #2466, opened 2025-12-18
Router factor out hotdeploy

predic8:
## Summary by CodeRabbit

* **New Features**
  * Centralized router lifecycle (init/start/stop/wait) and exposed rule management API.
  * Pluggable hot-deploy subsystem with real and no-op implementations and runtime enable/disable.
  * Runtime integrations: Kubernetes watcher, timer/HTTP client factories, and conditional JMX registration.
  * CLI startup refactor with structured exit signaling.

* **Bug Fixes**
  * Improved startup/shutdown sequencing, synchronization, error handling and logging.
  * Safer hot-deploy lifecycle and clearer CLI error reporting.

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/router/hotdeploy/DefaultHotDeployer.java
+++ b/core/src/main/java/com/predic8/membrane/core/router/hotdeploy/DefaultHotDeployer.java
@@ -13,9 +13,10 @@ public class DefaultHotDeployer implements HotDeployer {
     @GuardedBy("this")
     private HotDeploymentThread hdt;
 
-    private final Router router;
+    private Router router;
 
-    public DefaultHotDeployer(Router router) {
+    @Override
+    public void init(Router router) {
         this.router = router;
     }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/predic8/membrane/core/router/hotdeploy/DefaultHotDeployer.java
```diff
@@ -0,0 +1,61 @@
...
+
+import com.predic8.membrane.core.*;
+import com.predic8.membrane.core.config.spring.*;
+import org.slf4j.*;
+
+public class DefaultHotDeployer implements HotDeployer {
+
+    private static final Logger log = LoggerFactory.getLogger(DefaultHotDeployer.class.getName());
+
+    private HotDeploymentThread hdt;
```
rrayst, line 14: If access to `hdt` is protected by the `DefaultHotDeployer` lock, this should be documented "`@GuardedBy("this")`".

If this is the case, the `stop()` method should also acquire the lock before accessing `hdt`.

rrayst (changes requested):

rrayst:
Why is there a distinction between `router.hotDeployer instanceof NullHotDeployer` and `router.hotDeployer instanceof DefaultHotDeployer && ((DefaultHotDeployer)router.hotDeployer).hdt == null`?

In other words: why is `DefaultHotDeployer` not always running? This PR introduces additional state here.

rrayst:
maybe we could write an example test to test the `hotDeploy` feature?

PR #2513, opened 2025-12-24
YAML Refactor and support for global

predic8:
## Summary by CodeRabbit

* **New Features**
  * Added an openPorts flag to control whether proxies open ports automatically.
  * Unified and clearer startup logging for XML and YAML initialization flows.

* **Refactor**
  * Reorganized bean-related packages and refined bean registry/initialization behavior.
  * Adjusted proxy startup/initialization handling and lifecycle logging.

* **Chores**
  * Removed a global stopAll() public method.
  * Added human-readable string representations for bean container/definition objects.


annot/src/main/java/com/predic8/membrane/annot/beanregistry/BeanRegistryImplementation.java
```diff
@@ -79,8 +79,9 @@ public void handle(ChangeEvent changeEvent, boolean isLast) {
 
             if (!bd.isComponent() && observer.isActivatable(bd)) {
                 uidsToActivate.add(new UidAction(bd.getUid(), action));
+            } else if ("global".equals(bd.getKind())) {
```
rrayst: please avoid the special handling here.

`observer.isActivatable` could also return `true` to achieve the desired result.

rrayst (changes requested):

rrayst (approved):

PR #2514, opened 2025-12-25
Router refactor to beanfactory

predic8:
## Summary by CodeRabbit

* **New Features**
  * Lazy bean registration for on-demand component creation
  * Automatic rule reinitializer to retry and recover inactive routes
  * XML-based router bootstrap for loading router configuration

* **Bug Fixes**
  * Improved null-safety and startup robustness to avoid NPEs
  * Hardened error handling for malformed exchange/store responses
  * More diagnostic debug logging during startup and timer tasks

* **Documentation**
  * Roadmap updated with JMX exporter and lifecycle notes

* **Breaking**
  * Gatekeeper SSL interceptor has been removed

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/DefaultRouter.java
+++ b/core/src/main/java/com/predic8/membrane/core/DefaultRouter.java
@@ -125,7 +125,6 @@ public class DefaultRouter extends AbstractRouter implements ApplicationContextA
     /**
      * HotDeployer for changes on the XML configuration file. Does not cover YAML.
      * Not synchronized, since only modified during initialization
-     * Initialized with NullHotDeployer to avoid NPEs
      */
     private HotDeployer hotDeployer = new DefaultHotDeployer();
 
@@ -153,16 +152,14 @@ public class DefaultRouter extends AbstractRouter implements ApplicationContextA
         synchronized (lock) {
             if (initialized)
                 throw new IllegalStateException("Router already initialized.");
-        }
 
-        mainComponents.init();
+            mainComponents.init();
 
-        initProxies();
+            initProxies();
 
-        synchronized (lock) {
             initialized = true;
+            reinitializer = new RuleReinitializer(this); // Bean
         }
-        reinitializer = new RuleReinitializer(this); // Bean
     }
 
     /**
@@ -272,7 +269,7 @@ public class DefaultRouter extends AbstractRouter implements ApplicationContextA
 
     /**
      * @description A 'global' (per router) &lt;httpClientConfig&gt;. This instance is used everywhere
-     * a HTTP Client is used. Usually, in every specific place, you can still configure a local
+     * a HTTP Client is used. In every specific place, you should still be able to configure a local
      * &lt;httpClientConfig&gt; (with higher precedence compared to this global instance).
      */
     @MCChildElement()
@@ -426,11 +423,8 @@ public class DefaultRouter extends AbstractRouter implements ApplicationContextA
     @Override
     public void handleBeanEvent(BeanDefinitionChanged bdc, Object bean, Object oldBean) throws IOException {
         log.debug("Bean changed: type={} instance={}", bean.getClass().getSimpleName(), bean);
-        if (bean instanceof GlobalInterceptor) {
-            return;
-        }
-
         if (!(bean instanceof Proxy newProxy)) {
+            // should not happen, as handleBeanEvent() is only called for beans passing isActivatable().
             throw new IllegalArgumentException("Bean must be a Proxy instance, but got: " + bean.getClass().getName());
         }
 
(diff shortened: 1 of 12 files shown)


rrayst (dismissed):

annot/src/main/java/com/predic8/membrane/annot/beanregistry/BeanRegistryImplementation.java
```diff
@@ -85,38 +92,51 @@ public void handle(ChangeEvent changeEvent, boolean isLast) {
     }
 
     private void activationRun() {
-        Set<UidAction> uidsToRemove = new HashSet<>();
```
rrayst, line 88: produces a concurrent modification exception

predic8:
@coderabbitai review

annot/src/main/java/com/predic8/membrane/annot/beanregistry/BeanRegistryImplementation.java
```diff
@@ -85,38 +89,47 @@ public void handle(ChangeEvent changeEvent, boolean isLast) {
...
+            }
+
+            BeanDefinition def = bc.getDefinition();
+            Object oldBean;
+            Object newBean = null;
 
-                // e.g. inform router about new proxy
-                observer.handleBeanEvent(new BeanDefinitionChanged(uidAction.action, bc.getDefinition()), bean, getOldBean(uidAction.action, bc.getDefinition()));
+            try {
+                synchronized (bc) {
```
rrayst: why synchronize on a public monitor? this is bad practice.

and why synchronize on the bc's monitor, which is not documented and uses an AtomicReference internally???

annot/src/main/java/com/predic8/membrane/annot/beanregistry/BeanRegistryImplementation.java
```diff
@@ -13,40 +13,44 @@
...
+ */
 public class BeanRegistryImplementation implements BeanRegistry, BeanCollector {
 
     private static final Logger log = LoggerFactory.getLogger(BeanRegistryImplementation.class);
 
     private final BeanCacheObserver observer;
     private final Grammar grammar;
 
-    // uid -> bean
-    private final ConcurrentHashMap<String, Object> singletonBeans = new ConcurrentHashMap<>(); // Order is here not critical
```
rrayst, line 35: `singletonBeans` kept the *old* instance, e.g. in case of the modification of an `api`. it is required to implement `getOldBean` correctly.

rrayst (changes requested):

predic8:
@coderabbitai review

predic8:
@coderabbitai plz review

core/src/main/java/com/predic8/membrane/core/IRouter.java
```diff
@@ -0,0 +1,69 @@
...
+import com.predic8.membrane.core.resolver.*;
+import com.predic8.membrane.core.transport.*;
+import com.predic8.membrane.core.transport.http.*;
+import com.predic8.membrane.core.util.*;
+import org.springframework.context.*;
+
+import java.io.*;
+import java.util.*;
+
+public interface IRouter extends Lifecycle {
```
rrayst: The project does not use the `I` prefix for interfaces.

-> e.g. use `Router` and `RouterImpl`

rrayst (changes requested):

core/src/main/java/com/predic8/membrane/core/IRouter.java
predic8: Agree. It is just temporary. Plan:

- Make tests run again
- Router => DefaultRouter?,  IRouter => Router
- new TestRouter extends Router that opens ports but does not start HotDeployer,...
- HttpRouter => LightTestRouter that doesn't open ports

annot/src/main/java/com/predic8/membrane/annot/beanregistry/BeanContainer.java
```diff
@@ -14,24 +14,25 @@
...
 
+import java.util.concurrent.atomic.*;
+
 public class BeanContainer {
     private final BeanDefinition definition;
     /**
      * Constructed bean after initialization.
      */
-    private volatile Object singleton;
+    private final AtomicReference<Object> singleton = new AtomicReference<>();
```
rrayst, line 32: An `AtomicReference<T>`, on which only `T get()` and `void set(T)` are being called, behaves identically to a `volatile` field of type `T`. Therefore, this doesn't change anything.

core/src/main/java/com/predic8/membrane/core/proxies/SSLableProxy.java
```diff
@@ -46,7 +48,7 @@ public void init()  {
 			if (acmeCtx == null)
 				acmeCtx = new AcmeSSLContext(sslInboundParser, host, router.getHttpClientFactory(), router.getTimerManager());
 			setSslInboundContext(acmeCtx);
-			acmeCtx.init(router.getKubernetesClientFactory(), router.getHttpClientFactory());
+			acmeCtx.init(router);
```
rrayst, line 51: Eventually, the router should *not* be passed into all kinds of objects: Access to "global" functionality should happen only on a very limited basis.

rrayst (approved):
Approved. Before merging, we should check the test results.

PR #2525, opened 2026-01-02
Router: Refactoring move baselocation and isProduction to configuration

predic8:
…N` with static import and minor cleanup in methods

## Summary by CodeRabbit

* **Refactor**
  * Base location is now stored on the router's configuration object, unifying how resources and paths are resolved.

* **Breaking Changes**
  * Direct router-level baseLocation accessors removed — configuration API should be used.

* **Bug Fixes**
  * Resource, path, SSL and key lookups now consistently resolve relative to the configuration-provided base location.

* **Chores**
  * Tests and startup wiring updated to align with the new configuration-based approach.

### DIFF
--- a/core/src/test/java/com/predic8/membrane/integration/withoutinternet/MassivelyParallelTest.java
+++ b/core/src/test/java/com/predic8/membrane/integration/withoutinternet/MassivelyParallelTest.java
@@ -102,7 +102,7 @@ class MassivelyParallelTest {
                     es.submit(() -> job.accept(cdl));
                 }
                 es.shutdown();
-                if (!es.awaitTermination(60, SECONDS)) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/predic8/membrane/core/router/Configuration.java
```diff
@@ -140,4 +143,17 @@ public void setUriFactory(URIFactory uriFactory) {
     public URIFactory getUriFactory() {
         return uriFactory;
     }
+
+    /**
+     * @description Base location for the router's configuration file.
+     * @param baseLocation
+     */
+    @MCAttribute()
```
rrayst: why is `baseLocation` a `@MCAttribute`? it is the parent directory of the main config file.

rrayst (approved):

PR #2566, opened 2026-01-08
show load balancer node status (and failure reasons) in adminConsole

rrayst:


core/src/main/java/com/predic8/membrane/core/exchange/AbstractExchange.java
```diff
@@ -68,6 +70,9 @@ public abstract class AbstractExchange {
 
 	private ArrayList<Interceptor> interceptorStack = new ArrayList<>(10);
 
+	private int[] nodeStatusCodes;
```
predic8: Does that aspect belong to the Exchange?

core/src/main/java/com/predic8/membrane/core/interceptor/balancer/ExchangeNodeStatusTracker.java
```diff
@@ -0,0 +1,59 @@
+package com.predic8.membrane.core.interceptor.balancer;
+
+import static com.predic8.membrane.core.exchange.Exchange.TRACK_NODE_STATUS;
+import static java.lang.Boolean.TRUE;
+
+public class ExchangeNodeStatusTracker {
```
predic8, line 12: Add doc

PR #2628, opened 2026-01-17
Headerfilter Syntax for YAML

predic8:
## Summary by CodeRabbit

* **Refactor**
  * Replaced legacy header-filtering with a modular header filter and interceptor offering configurable include/exclude rules and an updated configuration API.

* **New Features**
  * Added a tutorial demonstrating header removal using the new header-filter.

* **Tests**
  * Added unit tests for rule matching and filter behavior; adapted existing tests to the new API and adjusted one expected error status from 500 to 400.

* **Chores**
  * Added/updated Apache 2.0 license headers in multiple test files.


predic8:
XML is not working at this stage. When we agree on the change we must add the modifications for XML.

PR #2671, opened 2026-01-23
Fix `BalancerUtil` to also return clusters and balancers defined in `global` or `chaindef`

christiangoerdes:


core/src/main/java/com/predic8/membrane/core/interceptor/balancer/BalancerUtil.java
```diff
@@ -16,56 +16,59 @@
...
 
 	public static List<LoadBalancingInterceptor> collectBalancers(Router router) {
 		ArrayList<LoadBalancingInterceptor> result = new ArrayList<>();
-		for (Proxy r : router.getRuleManager().getRules()) {
-			List<Interceptor> interceptors = r.getFlow();
-			if (interceptors != null)
-				for (Interceptor i : interceptors)
-					if (i instanceof LoadBalancingInterceptor)
-						result.add((LoadBalancingInterceptor)i);
+		for (List<Interceptor> flow : allFlows(router)) {
```
predic8: stream API

core/src/main/java/com/predic8/membrane/core/interceptor/balancer/BalancerUtil.java
```diff
@@ -16,56 +16,59 @@
...
+			if (flow == null) continue;
+			for (Interceptor i : flow)
+				if (i instanceof LoadBalancingInterceptor lbi)
+					result.add(lbi);
 		}
 		return result;
 	}
 
+	public static List<Cluster> collectClusters(Router router) {
+		ArrayList<Cluster> result = new ArrayList<>();
```
predic8: List

PR #2673, opened 2026-01-23
Refactor AccessLogInterceptorService to use `DateTimeFormatter` instead of `SimpleDateFormat`.

christiangoerdes:


core/src/main/java/com/predic8/membrane/core/interceptor/log/access/AccessLogInterceptorService.java
```diff
@@ -15,27 +15,35 @@
...
+import static com.predic8.membrane.core.util.TextUtil.escapeQuotes;
+import static java.time.ZoneId.systemDefault;
+import static java.time.format.DateTimeFormatter.ofPattern;
 
 public class AccessLogInterceptorService {
 
     private static final Logger log = LoggerFactory.getLogger(AccessLogInterceptorService.class);
 
-    private final SimpleDateFormat dateTimeFormat;
+    private final DateTimeFormatter dateTimeFormat;
```
predic8, line 49: comment

PR #2790, opened 2026-02-15
fix: compute URL expressions in the target after request flow

predic8:
…http client interceptor

## Summary by CodeRabbit

* **New Features**
  * Added URL encoding and path segment encoding functions for templating languages (Groovy, SpEL, JSONPath, XPath).
  * Added ability to change HTTP request methods dynamically and clear request bodies.
  * Support for expression-based URL template evaluation with configurable escaping strategies.

* **Bug Fixes**
  * Improved error handling for malformed and invalid URIs.
  * Enhanced validation of URI components with better security safeguards.

* **Improvements**
  * URL template expressions now evaluated at the appropriate stage in request processing.
  * Better control over character escaping in dynamic URL construction.

### DIFF
--- a/annot/src/main/java/com/predic8/membrane/annot/yaml/GenericYamlParser.java
+++ b/annot/src/main/java/com/predic8/membrane/annot/yaml/GenericYamlParser.java
@@ -45,6 +45,7 @@ import static com.predic8.membrane.annot.yaml.YamlParsingUtils.*;
 import static java.lang.reflect.Modifier.isAbstract;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.List.of;
+import static java.util.Locale.ROOT;
 import static java.util.UUID.randomUUID;
 
 public class GenericYamlParser {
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/predic8/membrane/core/proxies/Target.java
```diff
@@ -74,6 +77,44 @@ public Target(String host, int port) {
         setPort(port);
     }
 
+    public void applyModifications(Exchange exc, Router router) {
```
rrayst, line 114: can we rename `applyModifications` to `resolveExpressions`?

rrayst (changes requested):

core/src/main/java/com/predic8/membrane/core/util/URI.java
```diff
@@ -283,6 +282,23 @@ public String getPathWithQuery() {
         return r.toString();
     }
 
+    public URI resolve(URI relative) throws URISyntaxException {
```
rrayst, line 299: duplicate code: `ResolverMap.combine(baseUri, relativeUri)`.

rrayst (changes requested):

core/src/main/java/com/predic8/membrane/core/proxies/Target.java
predic8: Name is not great. We can change it. But it can also change the Method.

predic8:
@coderabbitai plz review

distribution/examples/configuration/apis.yaml
```diff
@@ -20,7 +20,10 @@ configuration:
   jmxRouterName: demo # Sets the JMX name for the router
 
   uriFactory:
-    allowIllegalCharacters: true # Allow non-RFC-compliant characters in URIs
+    # Allow non-RFC-compliant characters like {, } in URIs
+    # Attention: This may lead to security vulnerabilities! Use with care!
+    # - ${expression} in the URI will be evaluated
```
rrayst: behaviour has changed, please adjust comment

PR #2841, opened 2026-02-26
Fix XSLT relative include/import resolution by using ResolverMap URIResolver and correct systemId

christiangoerdes:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stylesheet path resolution and clearer error reporting for XSLT includes/imports.

* **Performance Improvements**
  * Reduced overhead when initializing XSLT transformers for better concurrent performance.

* **Examples**
  * Updated validation example to use inline field definitions for form validation.

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/interceptor/xslt/XSLTTransformer.java
+++ b/core/src/main/java/com/predic8/membrane/core/interceptor/xslt/XSLTTransformer.java
@@ -54,8 +54,9 @@ public class XSLTTransformer {
 		if (isNullOrEmpty(styleSheet))
 			t = fac.newTransformer();
 		else {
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/predic8/membrane/core/interceptor/xslt/XSLTTransformer.java
```diff
@@ -36,13 +36,30 @@ public XSLTTransformer(String styleSheet, final Router router, final int concurr
 		fac = TransformerFactory.newInstance();
 
 		this.styleSheet = styleSheet;
-		log.debug("using {} parallel transformer instances for {}",concurrency, styleSheet);
+
+		ResolverMap rr = router.getResolverMap();
+		String baseLocation = router.getConfiguration().getBaseLocation();
+
+		fac.setURIResolver((href, base) -> {
+			try {
```
predic8: extract

PR #2858, opened 2026-03-06
Add `cookie` to groovy context

christiangoerdes:
## Summary by CodeRabbit

* **New Features**
  * Scripts can now access HTTP cookies via "cookie" and "cookies" parameters to read individual cookie values.

* **Behavior**
  * Cookie parsing is performed lazily — cookies are parsed only when accessed to reduce overhead.

* **Tests**
  * Added tests for single and multiple cookie scenarios to validate parsing and parameter availability.

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/http/LazyCookieMap.java
+++ b/core/src/main/java/com/predic8/membrane/core/http/LazyCookieMap.java
@@ -17,11 +17,7 @@ package com.predic8.membrane.core.http;
 import com.predic8.membrane.core.http.cookie.Cookies;
 import com.predic8.membrane.core.http.cookie.MimeHeaders;
 
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
+import java.util.*;
 
 public class LazyCookieMap implements Map<String, String> {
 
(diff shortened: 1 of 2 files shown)


rrayst:
This PR would execute [relatively complex logic](https://github.com/membrane/api-gateway/blob/master/core/src/main/java/com/predic8/membrane/core/http/cookie/Cookies.java#L199) every time a groovy snippet gets executed. Do we want this?

predic8:
Lazy like in SpEL?

rrayst:
> Lazy like in SpEL?

yes, if possible.

core/src/main/java/com/predic8/membrane/core/lang/ScriptingUtils.java
```diff
@@ -139,4 +142,8 @@ public static Map<String, String> extractPathParameters(Exchange exchange) {
         // Add map to avoid a second parsing
         return emptyMap();
     }
+
+    private static Map<String, String> createCookieMap(Header header) {
+        return new LazyCookieMap(header);
```
predic8: inline

core/src/main/java/com/predic8/membrane/core/http/LazyCookieMap.java
```diff
@@ -0,0 +1,116 @@
...
+
+    private Map<String, String> map() {
+        if (delegate == null) {
+            delegate = parseCookies(header);
+        }
+        return delegate;
+    }
+
+    private static Map<String, String> parseCookies(Header header) {
+        var parsed = new HashMap<String, String>();
```
predic8: Linked

predic8 (approved):

PR #2873, opened 2026-03-16
Chain description

predic8:
## Summary by CodeRabbit

* **New Features**
  * Chain interceptors now expose user-facing metadata: a display name, a short description, and a detailed (HTML-formatted) description listing contained interceptors to improve UI labels and help/documentation.


core/src/main/java/com/predic8/membrane/core/interceptor/chain/ChainInterceptor.java
```diff
@@ -25,6 +25,21 @@
 
 /**
  * @description A Chain groups multiple interceptors into reusable components, reducing redundancy in API configurations.
+ *  @yaml <pre><code>
+ *  components:
+ *    my-chain:
```
christiangoerdes, line 30: `chain:`  is missing

PR #2886, opened 2026-03-18
Proxy: Fix URI for CONNECT Requests

predic8:
## Summary by CodeRabbit

* **New Features**
  * Exposes proxy CONNECT tunneling status to callers.

* **Bug Fixes**
  * Clearer proxy handshake error messages.
  * Improved handling and error reporting for malformed destination URLs.
  * Better host:port parsing (including IPv6 cases) and stricter non-null contract for host formatting.

* **Refactor**
  * Reorganized HTTP request, host/port resolution, and proxy handling flow.
  * Stream setup and authorization header flow modernized.

* **Tests**
  * Updated and modernized tests; added license header and new IPv6 coverage.


core/src/main/java/com/predic8/membrane/core/transport/http/Connection.java
```diff
@@ -77,6 +77,9 @@ public class Connection implements Closeable, MessageObserver, NonRelevantBodyOb
 	private Exchange exchange;
 	private boolean keepAttachedToExchange;
 
+	/**
+	 * If the connection is tunneled through a proxy
```
rrayst: and connection to the real backend uses TLS

PR #2959, opened 2026-05-27
Add support for YAML hot deployment and configuration reloading

christiangoerdes:
## Summary by CodeRabbit

* **New Features**
  * YAML-based hot reloading to update configuration dynamically without restarting.
  * Hot deployment monitors YAML config files, can be enabled/disabled, and validates changes before applying with automatic recovery on failure.
  * Router exposes YAML config location and tracked files for monitoring.

* **Refactor**
  * Simplified YAML bootstrap and startup flow for clearer startup/shutdown logging and more reliable lifecycle handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/membrane/api-gateway/pull/2959?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


core/src/main/java/com/predic8/membrane/core/router/hotdeploy/YamlHotDeploymentThread.java
```diff
@@ -0,0 +1,124 @@
...
+                    if (isInterrupted() || !hotDeployer.isEnabled()) {
+                        log.debug("YAML Hot Deployment Thread interrupted.");
+                        return;
+                    }
+
+                    //noinspection BusyWait
+                    sleep(1000);
+                }
+
+                log.debug("yaml configuration changed.");
```
predic8: info

core/src/main/java/com/predic8/membrane/core/router/DefaultRouter.java
```diff
@@ -454,6 +474,92 @@ public RuleReinitializer getReinitializer() {
         return reinitializer;
     }
 
+    public void setYamlConfiguration(String yamlConfigurationLocation, List<Path> trackedFiles) {
```
predic8: Move with tracedFiles into a separate class e.g. YamlConfigurationReloader

predic8 (dismissed):

core/src/main/java/com/predic8/membrane/core/router/DefaultRouter.java
```diff
@@ -454,6 +461,68 @@ public RuleReinitializer getReinitializer() {
...
+        }
+    }
+
+    public void stopRuntimeForReload() {
+        getRegistry().getBean(KubernetesWatcher.class).ifPresent(KubernetesWatcher::stop);
+
+        if (reinitializer != null)
+            reinitializer.stop();
+        if (mainComponents.getTransport() != null)
+            mainComponents.getTransport().closeAll(false);
```
rrayst: This does not wait for old Exchanges to be completed. This will cause issues, e.g. when the old Membrane writes to access logs/Exchange stores.

core/src/main/java/com/predic8/membrane/core/router/hotdeploy/YamlRouterReloader.java
```diff
@@ -0,0 +1,121 @@
...
+            return true;
+        } catch (Exception e) {
+            logReloadFailure(e);
+            if (!runtimeStopped) {
+                log.info("Keeping the previous YAML runtime because reload validation failed before shutdown.");
+                return true;
+            }
+            try {
+                router.stopRuntimeForReload();
+                router.disposeRuntime();
```
rrayst: Once a reloading has failed, the Membrane process terminates. -> No recovery is possible.

predic8:
# Branch review: yaml-hot-deployment vs master

**Verdict:** Nearly there — one blocker (resource leak on each successful reload) and one question mark (dropped `registry.start()`) before merge.

The branch wires YAML hot deployment end-to-end: a `YamlHotDeploymentThread` polls tracked files for changes, a `YamlRouterReloader` owns the validate-stop-reload-start sequence, and `DefaultRouter` gains `markReloading`/`clearReloading` so `waitFor()` holds the process alive during a reload. The refactoring of `RouterCLI` and `DefaultHotDeployer` is clean. Error paths (validation failure, post-shutdown failure) are both tested.

---

## Blockers

**`DefaultRouter.stopRuntimeForReload()` — old `TimerManager` and registry abandoned without cleanup.**

In the successful reload sequence — `stopRuntimeForReload()` → `resetRuntime()` → `loadIntoRouter()` → `start()` — the old `mainComponents` is silently replaced. `resetRuntime()` does **not** call `mainComponents.getTimerManager().shutdown()` (cancels the `java.util.Timer` and any pending tasks) or `closeRegistryIfSupported()` before discarding them. `BeanRegistryImplementation.close()` runs all `@PreDestroy` callbacks; skipping it means any interceptor or service with a `@PreDestroy` teardown (open sockets, background threads, connection pools) leaks resources on every hot reload.

The failure-cleanup path already calls `disposeRuntime()` (which does both). The normal `stop()` path also calls both. Only the success path is missing them.

> **Prompt to fix:** Add `mainComponents.getTimerManager().shutdown()` and `closeRegistryIfSupported()` to `stopRuntimeForReload()`, before closing the transport — or extract a `shutdownRuntimeComponents()` helper and call it from both `stopRuntimeForReload()` and `disposeRuntime()`. Make the success and failure cleanup paths symmetric.

---

## Improvements

**`YamlRouterBootstrap.loadIntoRouter()` — `registry.start()` dropped without explanation.**

The original `RouterCLI.initRouterByYAML()` called `reg.finishStaticConfiguration()` then `reg.start()`. `BeanRegistryImplementation.start()` eagerly calls `bc.getOrCreate(this, grammar)` on every `BeanContainer`, instantiating all beans up front. `loadIntoRouter()` omits `registry.start()`, so only explicitly-resolved beans (e.g., the `Configuration`) are created at parse time.

If any bean is a `BeanCacheObserver` or background service that was previously alive because `start()` instantiated it but is never explicitly resolved downstream, this is a silent regression. Unit tests pass because the test fixture is a single simple `internal` proxy; a config with observers or JMX beans might not.

> **Prompt to fix:** Check what `BeanRegistryImplementation.start()` bootstraps that is not covered by `initProxies()` / explicit bean resolution. Either restore `registry.start()` inside `loadIntoRouter()`, or add a comment explaining why it is intentionally omitted (e.g., "proxies are created during `initProxies()` via bean resolution; nothing requires eager start here").

**`YamlHotDeploymentTest` — success path not covered.**

Both tests verify failure recovery. There is no test that writes a valid new config and asserts the router picks up the new rules. This is the hot path exercised by every real-world hot reload.

> **Prompt to fix:** Add a test `shouldReloadSuccessfully`: write a `first` config, start the router, overwrite the file with a `second` config, call `reloader.reload()`, assert it returns `true`, and assert the rule manager now contains `"second"`.

---

## Nitpicks

- **`YamlHotDeploymentThread` — exit log is wrong for non-interrupt exits.** `"YAML Hot Deployment Thread interrupted."` is printed on all three exit paths: interrupt signal, `reload()` returning `false`, and `enabled` becoming `false`. Only the first is an interrupt; the other two are graceful stops. Use separate log messages so the distinction shows up in logs.

- **`DefaultRouter.markReloading()` — `notifyAll()` is vacuous here.** When `reloading` is set to `true`, the notification wakes `waitFor()` callers, which immediately see `reloading == true` and go back to sleep. The meaningful notification is in `clearReloading()`. The call in `markReloading()` is harmless but noisy — consider removing it.

rrayst (changes requested):

rrayst (approved):

PR #2963, opened 2026-05-28
add JsonRPCProtection plugin

christiangoerdes:
## Summary by CodeRabbit

* **New Features**
  * JSON-RPC protection interceptor with request validation, explicit allow/deny rules, and pattern-based method matching
  * Batch handling configurable (enable/disable, max size, empty/oversize rejection)
  * Method-specific params validation via regex→schema mappings (YAML/XML support, first-match wins)
  * Improved error responses and a now-public JSON-RPC request helper

* **Tests**
  * Expanded tests covering rule precedence, rejections, batch behavior/limits, param validation, and invalid configs

[Review Change Stack](https://app.coderabbit.ai/change-stack/membrane/api-gateway/pull/2963?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


core/src/main/java/com/predic8/membrane/core/interceptor/json/rpc/JsonRPCProtectionInterceptor.java
```diff
@@ -0,0 +1,199 @@
...
+ *
+ * @yaml
+ * <pre><code>
+ * - jsonRPCProtection:
+ *     batch:
+ *       enabled: true
+ *       maxSize: 50
+ *     methods:
+ *       - allow: "^rpc\\.(health|echo)$"
+ *       - deny: "^rpc\\..*$"
```
predic8: deny: *

core/src/main/java/com/predic8/membrane/core/util/allowdeny/Allow.java
```diff
@@ -0,0 +1,20 @@
+package com.predic8.membrane.core.util.allowdeny;
```
predic8: util/config/allowdeny

predic8 (changes requested):

annot/src/main/java/com/predic8/membrane/annot/yaml/error/LineYamlErrorRenderer.java
```diff
@@ -283,31 +283,40 @@ private static int getIndentation(String line) {
...
-            // Last segment is object key like .field
-            return jsonPath.substring(0, lastDot);
+    private static String getLastSegment(String jsonPath) {
+        int start = findLastSegmentStart(jsonPath);
+        String segment = jsonPath.substring(start);
+
+        if (segment.startsWith(".")) {
+            return segment.substring(1);
+        }
+
```
predic8, line 302: Doc sample ...

predic8: extract

annot/src/main/java/com/predic8/membrane/annot/yaml/error/LineYamlErrorRenderer.java
```diff
@@ -283,31 +283,40 @@ private static int getIndentation(String line) {
...
+        if (segment.startsWith("['") && segment.endsWith("']")) {
+            return segment.substring(2, segment.length() - 2)
```
predic8: extract

annot/src/main/java/com/predic8/membrane/annot/yaml/error/LineYamlErrorRenderer.java
```diff
@@ -283,31 +283,40 @@ private static int getIndentation(String line) {
...
+                    .replace("\\'", "'")
+                    .replace("\\\\", "\\");
         }
+
+        if (segment.startsWith("[") && segment.endsWith("]")) {
+            return segment.substring(1, segment.length() - 1);
```
predic8, line 308: extract

annot/src/main/java/com/predic8/membrane/annot/yaml/parsing/binding/ScalarValueConverter.java
```diff
@@ -92,12 +94,42 @@ private Object coerceNonTextual(ParsingContext<?> ctx, Method setter, JsonNode n
...
             return node.isBoolean() ? node.booleanValue() : parseBoolean(node.asText());
         if (wanted.equals(Map.class) && setter != null && hasOtherAttributes(setter))
-            return Map.of(key, node.asText());
+            return Map.of(key, convertAnySetterValue(ctx, setter, node, key));
         if (setter != null && isReferenceAttribute(setter))
             return resolveRegistryReference(ctx, node.asText(), key);
         throw unsupported(wanted, key, node);
     }
 
+    private Object convertAnySetterValue(ParsingContext<?> ctx, Method setter, JsonNode node, String key) {
```
predic8, line 109: doc

annot/src/main/java/com/predic8/membrane/annot/yaml/ParsingContext.java
```diff
@@ -84,6 +88,16 @@ public String getPath() {
         return path;
     }
 
+    private static String toJsonPathProperty(String property) {
```
predic8, line 98: docs

core/src/main/java/com/predic8/membrane/core/interceptor/json/rpc/JsonRPCMethodDefinitions.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Map;
+
+/**
+ * @description
+ * <p>Maps JSON-RPC method names to their schema validation definitions.</p>
+ *
+ * <p>In YAML, the entries are written as a map under <code>schemaValidation.methods</code>.</p>
+ */
+@MCElement(name = "methods", component = false, id = "json-rpc-method-definitions")
+public class JsonRPCMethodDefinitions {
```
predic8: JsonRPCMethodsDefinitions

core/src/main/java/com/predic8/membrane/core/interceptor/json/rpc/JsonRPCSchemas.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * @description
+ * <p>Defines the JSON Schema validation rules for one exact JSON-RPC method name.</p>
+ *
+ * <p>Use <code>params</code> to validate the request payload and <code>response</code>
+ * to validate successful upstream responses.</p>
+ */
+@MCElement(name = "method", component = false, id = "json-rpc-method-schema")
+public class JsonRPCSchemas {
```
predic8: rename

core/src/main/java/com/predic8/membrane/core/interceptor/json/rpc/JsonRPCSchemaValidation.java
```diff
@@ -0,0 +1,345 @@
...
+            throw new ConfigurationException("Cannot initialize JSON-RPC schema validation without resolver context.");
+        }
+
+        SchemaRegistry registry = createSchemaRegistry(resolver);
+        errorSchema = resolveErrorSchema(registry, resolver, uriFactory, beanBaseLocation);
+
+        Map<String, Schema> resolvedParamSchemas = new LinkedHashMap<>();
+        Map<String, Schema> resolvedResponseSchemas = new LinkedHashMap<>();
+
+        for (Map.Entry<String, JsonRPCSchemas> entry : methods.getMethods().entrySet()) {
```
predic8: extract method

predic8 (changes requested):

distribution/tutorials/json/30-JSON-RPC-Protection.yaml
```diff
@@ -0,0 +1,88 @@
...
+api:
+  port: 2000
+  flow:
+      - jsonRPCProtection:
+          batch:
+            enabled: true
+            maxSize: 2 # At most two calls per batch
+          methods:
+            - allow: '^rpc\.(health|echo)$'
+            - deny: '.*' # Default deny for everything else
```
predic8: rm Default

distribution/tutorials/security/json-rpc/20-JSON-RPC-Protection-with-Schema-Validation.yaml
```diff
@@ -0,0 +1,88 @@
+# yaml-language-server: $schema=https://www.membrane-api.io/v7.2.2.json
```
predic8, line 1: 2 tutorials one simple without schema/regex pattern

predic8 (changes requested):
Dead code — remove. JsonRPCValidator.validate(String) (JsonRPCValidator.java:54) has no callers (prod or test). Same for JsonRPCSchemaValidation.isEmpty() and the hasRequestValidation() it depends on (JsonRPCSchemaValidation.java:121,137). Fix: delete all three.

Duplicated payload-type logic. JsonRPCProtectionInterceptor.payloadType (:232) and JsonRPCValidator.getPayloadType (:336) are byte-for-byte identical. Fix: keep one (e.g. a static on PayloadType) and reuse.

core/src/main/java/com/predic8/membrane/core/interceptor/json/rpc/JsonRPCParams.java
```diff
@@ -0,0 +1,136 @@
...
+@MCElement(name = "params", component = false)
+public class JsonRPCParams {
+
+    private Map<String, String> params = new LinkedHashMap<>();
+    private List<CompiledSchema> schemas = List.of();
+
+    @MCOtherAttributes
+    public void setParams(Map<String, String> params) {
+        this.params = params == null ? new LinkedHashMap<>() : new LinkedHashMap<>(params);
+    }
```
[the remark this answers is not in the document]
rrayst: @predic8 is this an issue?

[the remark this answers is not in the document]
rrayst: ok, outdated and addressed.

rrayst (dismissed):

distribution/tutorials/security/json-rpc/20-JSON-RPC-Protection-with-Schema-Validation.yaml
```diff
@@ -0,0 +1,88 @@
...
+
+api:
+  port: 2000
+  flow:
+      - jsonRPCProtection:
+          batch:
+            enabled: true
+            maxSize: 2 # At most two calls per batch
+          methods:
+            - allow: '^rpc\.(health|echo)$'
```
predic8: Write sample with get* allow pattern to allow read only

predic8 (changes requested):

predic8:
# Branch review: json-rpc-protection vs master

**Verdict:** Solid feature, one correctness bug to fix before merge.

This branch adds a `jsonRPCProtection` interceptor for validating JSON-RPC requests and responses: allow/deny rules on method names, batch size limiting, and JSON Schema validation of params/responses/errors. It also adds infrastructure work in `annot` to support `@MCOtherAttributes` with typed map values (enabling the `methods: { "rpc.echo": {...} }` YAML shape) and improved JSONPath rendering for dotted keys. Test coverage is thorough.

## Build & tests
- `mvn -pl core test -Duser.language=en -Duser.country=US` → **PASS**, 3430 tests, 0 failures.
- `mvn -pl annot test -Duser.language=en -Duser.country=US` → **1 pre-existing failure** (`SpringConfigXSDErrorsTest.mcElementNameMissing`, locale-sensitive javac message matching; also fails on master with this JVM). Not a branch regression.
- `distribution` IT tests not run — integration tests require a built distribution.

---

## Blockers

- **`JsonRPCValidator.java:173-179` — Notification + combined error+response validation rejects valid backend responses.**

  When both `schemaValidation.error` and `schemaValidation.methods[x].response` are configured, a notification request (no `id` field) passes through request validation with an empty `methodsById` map. `createResponseValidationContext` keeps a non-null context because `hasErrorValidation()` is true. When the backend's success response arrives, `validateSingleResponse` passes the `!isError()` branch, then hits `hasMethodResponseValidation() == true`, and falls through to `context.methodFor(response.getId())` which returns `null` because no method was remembered — producing a 500 "response id does not match any request" rejection for a perfectly valid exchange.

  No test covers the combined-validation + notification path. The existing test `"notification responses are ignored when only success schemas are configured"` passes because `hasErrorValidation()` is false there, so `createResponseValidationContext` returns `null` and `validateResponse` exits early.

  *Fix:* In `validateSingleResponse` (`JsonRPCValidator.java:173`), extend the early-return guard:
  ```java
  if (!schemaValidation.hasMethodResponseValidation() || !context.expectsResponses()) {
      return null;
  }
  ```
  Then add a test case: notification + both error and response schemas configured + success response → `CONTINUE`.

---

## Improvements

- **`LineYamlErrorRenderer.java:332` — `findLastSegmentStart` finds dots inside `['rpc.echo']` brackets.**

  `lastIndexOf('.')` is used to find the last segment boundary, but dots inside a `['foo.bar']`-style segment (produced by the new `toJsonPathProperty`) will be found first. For a path like `$.api.methods['rpc.echo']`, the method returns offset 18 (the dot inside `rpc.echo`) instead of offset 13 (the `[`). `getParentPath` would produce `"$.api.methods['rpc."` and `getLastSegment` would return `"echo']"`.

  This only triggers when `key == null` in `getFormattedReport` (i.e., the error is AT the map entry itself, not inside it — for example, `'rpc.echo': "wrong type"`). The new test covers the `key != null` case (error inside the entry), which is why it passes. The old code had the same logic, but `toJsonPathProperty` makes this case newly reachable.

  *Fix:* In `findLastSegmentStart`, scan backwards while tracking bracket depth:
  ```java
  private static int findLastSegmentStart(String path) {
      int depth = 0;
      for (int i = path.length() - 1; i >= 0; i--) {
          char c = path.charAt(i);
          if (c == ']') depth++;
          else if (c == '[') depth--;
          if (depth == 0 && (c == '.' || c == '[')) return i;
      }
      throw new IllegalArgumentException("Cannot determine parent path of: " + path);
  }
  ```
  Add a test: `getLastSegment("$.foo['bar.baz']")` == `"bar.baz"` and `getParentPath` == `"$.foo"`.

- **`JsonRPCProtectionInterceptor.java:185` — `setSchemaValidation` missing `@description`.**

  All other `@MCChildElement` setters in this class are documented; this one has no Javadoc at all. *Fix:* Add `/** @description Configures JSON Schema validation for params, responses, and error objects. */`.

- **`@MCChildElement(order=4)` gap** (`JsonRPCProtectionInterceptor.java:185`): Orders used are 0, 1, 4. Change to `order=2`.

- **`JsonRPCSchemaValidation.isEmpty()` is dead code** (`JsonRPCSchemaValidation.java:127`): Never called. Remove it.

---

## Nitpicks

- `JsonRPCValidator.validate(String body)` (line 51) — public wrapper around `validateRequest(body).error()` that nothing calls. Remove or make package-private.
- Tutorial `10-JSON-RPC-Allow-Deny-and-Batch-Validation.yaml:44-46` — `allow: ping` works correctly (Java `Pattern.matches` is full-string), but looks inconsistent with tutorial 20's explicit `'^...$'` anchors. Consider adding a comment or using anchors throughout.

predic8:
## AI instructions to fix the review findings

Self-contained prompt — paste into Claude Code on the `json-rpc-protection` branch:

---

Fix the following review findings on this branch. For each fix, also add or update a test that covers the corrected behaviour. Run `mvn -pl core test -Duser.language=en -Duser.country=US` and `mvn -pl annot test -Duser.language=en -Duser.country=US` after all changes and confirm both pass before reporting done.

### 1 · BLOCKER — notification + combined validation rejects valid responses

`core/src/main/java/com/predic8/membrane/core/interceptor/json/rpc/JsonRPCValidator.java`

In `validateSingleResponse`, the guard at line ~173 reads:
```java
if (!schemaValidation.hasMethodResponseValidation()) {
    return null;
}
```
Change it to:
```java
if (!schemaValidation.hasMethodResponseValidation() || !context.expectsResponses()) {
    return null;
}
```

**Why:** When both `schemaValidation.error` and a per-method `response` schema are configured, a notification request (no `id`) produces an empty `methodsById` map. The context is kept non-null because `hasErrorValidation()` is true. When the backend returns a success response, `context.methodFor(id)` always returns null → false rejection with "response id does not match any request."

**Test to add** in `JsonRPCProtectionInterceptorTest` (in `responseCases()`):
- config: both `ERROR_INLINE_CONFIG` and `RESPONSE_INLINE_CONFIG` merged (error schema + rpc.inline response schema)
- request body: `{"jsonrpc":"2.0","method":"rpc.inline"}` (notification — no `id`)
- response body: `{"jsonrpc":"2.0","id":1,"result":{"ok":true}}`
- expected outcome: `CONTINUE`

---

### 2 · IMPROVEMENT — `findLastSegmentStart` breaks on dots inside `['rpc.echo']`

`annot/src/main/java/com/predic8/membrane/annot/yaml/error/LineYamlErrorRenderer.java`

Replace the `findLastSegmentStart` method body with a bracket-depth scan so interior dots are ignored:

```java
private static int findLastSegmentStart(String jsonPath) {
    int depth = 0;
    for (int i = jsonPath.length() - 1; i >= 0; i--) {
        char c = jsonPath.charAt(i);
        if (c == ']') depth++;
        else if (c == '[') depth--;
        if (depth == 0 && (c == '.' || c == '[')) return i;
    }
    throw new IllegalArgumentException("Cannot determine parent path of: " + jsonPath);
}
```

**Why:** `lastIndexOf('.')` finds the dot inside `['rpc.echo']` before the opening bracket, producing a malformed parent path when `key == null` and the path ends with a quoted bracketed segment. The new `toJsonPathProperty` method makes this case reachable.

**Tests to add** in `YAMLParsingErrorTest` (or a dedicated `LineYamlErrorRendererTest`):

- `getLastSegment("$.foo['bar.baz']")` == `"bar.baz"`
- `getParentPath("$.foo['bar.baz']")` == `"$.foo"`
- `getLastSegment("$.foo['bar\\'baz']")` == `"bar'baz"`
- `getParentPath("$.foo.child")` == `"$.foo"` (existing behaviour unchanged)

---

### 3 · IMPROVEMENT — `setSchemaValidation` missing `@description`

`core/src/main/java/com/predic8/membrane/core/interceptor/json/rpc/JsonRPCProtectionInterceptor.java`

Add Javadoc before `setSchemaValidation`:
```java
/**
 * @description Configures JSON Schema validation for request params, successful responses, and error objects.
 */
@MCChildElement(order = 4)
public void setSchemaValidation(...) {
```

---

### 4 · IMPROVEMENT — `@MCChildElement(order=4)` gap

Same file as above. Change `order = 4` to `order = 2` on `setSchemaValidation`. The three child elements should use consecutive orders 0, 1, 2.

---

### 5 · IMPROVEMENT — `JsonRPCSchemaValidation.isEmpty()` is dead code

`core/src/main/java/com/predic8/membrane/core/interceptor/json/rpc/JsonRPCSchemaValidation.java`

Remove the `isEmpty()` method — it is never called.

---

### 6 · NITPICK — `JsonRPCValidator.validate(String body)` is an unused public method

`core/src/main/java/com/predic8/membrane/core/interceptor/json/rpc/JsonRPCValidator.java`

Remove the `validate(String body)` method (it only delegates to `validateRequest(body).error()` and nothing calls it).

---

### 7 · NITPICK — tutorial allow/deny pattern style

`distribution/tutorials/security/json-rpc/10-JSON-RPC-Allow-Deny-and-Batch-Validation.yaml`

The patterns `ping` and `sum` are correct (Java `Pattern.matches` does full-string matching) but look inconsistent with tutorial 20's explicit `'^...$'` style. Add a comment above the `methods:` block, e.g.:

```yaml
# Patterns use Java regex full-string matching (like ^...$).
# Use .* for wildcards, e.g. deny: .* blocks everything not matched above.
```

predic8 (approved):

PR #2997, opened 2026-06-23
ResolverMap.combine: Fix combine for paths linek ?foo=http://

predic8:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved how URL/path combination detects URI schemes, preventing incorrect scheme handling when resolving parent and child locations (including cases with queries and already-schemed child paths).

* **Tests**
  * Added new JUnit 5 coverage to validate URL resolution behavior across multiple scenarios and to verify the scheme-detection logic, including edge cases for false positives and varied scheme formats.


core/src/main/java/com/predic8/membrane/core/resolver/ResolverMap.java
```diff
@@ -111,9 +117,20 @@ private static String combineInternal(URIFactory factory,String... locations) {
...
+    private static final Pattern SCHEME = Pattern.compile("^[a-zA-Z][a-zA-Z0-9+.-]*:[/\\\\]");
+
+    private static boolean hasScheme(String location) {
+        return SCHEME.matcher(location).find();
+    }
+
     private static String combineInternal2(URIFactory uriFactory, String[] locations, String relativeChild, String parent) {
 
-        if (relativeChild.contains(":/") || relativeChild.contains(":\\") || parent == null || parent.isEmpty())
+        if (hasScheme(relativeChild) || parent == null || parent.isEmpty())
```
rrayst, line 133: `:\` is not included in the check anymore?

predic8: Yes — :\ is still part of the check. It's just hidden behind two escaping layers. The pattern in ResolverMap.java:125:

Pattern.compile("^[a-zA-Z][a-zA-Z0-9+.-]*:[/\\\\]");

I added a new test to verify it.

rrayst (approved):

PR #3033, opened 2026-07-03
OpenAPI Validator: sanitize request method logging and remove method from problem details

predic8:
…ssignment

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OpenAPI validation failure logging by truncating and masking non-printable characters in request method and URI for request/response paths and “no matching API found” details.
  * Sanitized related validation error messages (including method/path details) and improved correlation-id header masking to prevent raw non-printable values from appearing.
* **Tests**
  * Updated unit tests to reflect the renamed non-printable masking helper.


core/src/main/java/com/predic8/membrane/core/openapi/serviceproxy/OpenAPIInterceptor.java
```diff
@@ -105,34 +105,36 @@ public APIProxy getApiProxy() {
...
+        var basePath = getMatchingBasePath(exc);
         // No matching API found
         if (basePath == null) {
             // Do not log: 404 is too common
             user(false, getDisplayName())
                     .title("No matching API found!")
                     .status(404)
                     .addSubSee("not-found")
-                    .detail("There is no API on the path %s deployed. Please check the path.".formatted(exc.getOriginalRequestUri()))
+                    .detail("There is no API on the path %s deployed. Please check the path.".formatted(truncateAndMaskNonPrintable(exc.getOriginalRequestUri())))
```
rrayst, line 116: shouldn't 'escaping/making printable of the details(...) parameter' be a concern of the `ProblemDetails` message? Grepping the code for `.detail(` shows several occurences of "attacker-controlled" string parts in the parameter.

rrayst: or maybe change the signature to `.detail(String messageWithPlaceholders, Object... placeholders)` ?

rrayst (changes requested):

PR #3039, opened 2026-07-04
fix: duplicate 'Content-Length' should return an error

rrayst:
this enforces RFC 9112 Section 6.3

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened `Content-Length` handling by strictly validating all occurrences (including comma-separated tokens) and rejecting mismatches with a malformed-header failure.
  * Invalid framing now results in a `400 Bad Request` with connection close; backend invalid responses now return `502 Bad Gateway` (invalid framing).
* **Tests**
  * Added coverage for missing/zero/matching repeated or comma-delimited `Content-Length` values and for invalid formats/values.


core/src/main/java/com/predic8/membrane/core/http/Header.java
```diff
@@ -348,10 +348,55 @@ public boolean isChunked() {
...
+            }
+        }
+        return length == null ? -1 : length;
+    }
+
+    private static long parseContentLengthValue(String token, String rawValue) {
+        String trimmed = token.trim();
+        boolean digitsOnly = !trimmed.isEmpty();
+        for (int i = 0; i < trimmed.length(); i++) {
+            char c = trimmed.charAt(i);
```
predic8: May be more  compact:

trimmed.chars().allMatch(c -> c >= '0' && c <= '9')

core/src/main/java/com/predic8/membrane/core/transport/http/HttpServerHandler.java
```diff
@@ -146,6 +146,11 @@ public void run() {
             log.debug("Client connection terminated before first line was read. Line so far: {}", getLineMaskedAndTruncated(e));
         } catch (EOFWhileReadingLineException e) {
             log.debug("Client connection terminated while reading header line: {}", getLineMaskedAndTruncated(e));
+        } catch (MalformedHeaderException e) {
```
predic8, line 150: What about a backend sending two content-len?

core/src/main/java/com/predic8/membrane/core/transport/http/HttpServerHandler.java
```diff
@@ -367,6 +372,17 @@ private void updateThreadName(boolean fromConnection) {
         currentThread().setName(DEFAULT_THREAD_NAME);
     }
 
+    private void respondWithBadRequestAndClose(String message) {
+        try {
```
predic8, line 377: Return problem JSON, e.g.:

```
private void respondWithBadRequestAndClose(String message) {
    try {
        Response response = user(transport.getRouter().getConfiguration().isProduction(), "http-server-handler")
                .addSubType("invalid-framing")
                .detail(message)
                .build();
        response.getHeader().setConnection(Header.CLOSE);
        response.write(srcOut, false);
        srcOut.flush();
    } catch (Exception e) {
        log.debug("Could not send 400 response for request with invalid framing.", e);
    }
}
```

predic8 (changes requested):

core/src/main/java/com/predic8/membrane/core/http/Header.java
rrayst: Done in c505f94 — switched to `trimmed.chars().allMatch(c -> c >= '0' && c <= '9')`. Kept an explicit `!isEmpty()` guard, since `allMatch` is vacuously true on an empty string and we still want to reject an empty value.

core/src/main/java/com/predic8/membrane/core/transport/http/HttpServerHandler.java
rrayst: Good catch. A backend returning conflicting `Content-Length` is parsed at a different call site than the client request: the request is parsed in `HttpServerHandler` (caught here → 400 + close), while the backend response is parsed inside `HTTPClientInterceptor.handleRequest` via `httpClient.call(...)`. So no origin tagging on the exception is needed — the two are separable by where they're caught.

Fixed in ec744457b: `HTTPClientInterceptor` now catches `MalformedHeaderException` before the generic handler and returns a **502 Bad Gateway** problem+json (`ProblemDetails.gateway(...)`, subtype `invalid-framing`) instead of the previous generic 500.

- request with duplicate `Content-Length` → 400 (unchanged)
- backend response with duplicate `Content-Length` → 502 `invalid-framing` (new)

Note: the HTTP/2 path (`Http2ExchangeHandler`) still falls back to a generic 500 — I'll address that together with CodeRabbit's cross-call-site comment.

predic8 (approved):

PR #3072, opened 2026-07-08
Hardened XML parsers and validators

predic8:
…bilities

## Summary by CodeRabbit

* **New Features**
  * Added hardened XML/SOAP handling for schema validation and XML parsing, including hardened schema/validator processing and XXE-blocking parsers/factories.
* **Bug Fixes**
  * Prevents external DTD/entity retrieval during SOAP analysis, XML protection, XML parsing, and schema validation.
  * Improves DOCTYPE/external subset detection; when DTD removal is enabled, DOCTYPE is stripped and external endpoints are not contacted.
* **Tests**
  * Added security-focused tests to verify no external DTD fetches across the affected flows.
* **Refactor**
  * Standardized hardened XML reader/input factory usage across components.


core/src/main/java/com/predic8/membrane/core/interceptor/soap/SoapOperationExtractor.java
```diff
@@ -93,11 +93,8 @@ private void moveToTag(XMLStreamReader reader) throws XMLStreamException {
 		}
 	}
 
-	private XMLStreamReader getReader(Exchange exc) throws XMLStreamException,
-	FactoryConfigurationError, IOException {
-		synchronized (xmlInputFactory) {
-			return xmlInputFactory.createXMLStreamReader(xopr.reconstituteIfNecessary(exc.getRequest()));
-		}
+	private XMLStreamReader getReader(Exchange exc) throws XMLStreamException, IOException {
+		return XMLInputFactoryFactory.inputFactory().createXMLStreamReader(xopr.reconstituteIfNecessary(exc.getRequest()));
```
rrayst: can we cache the `XMLInputFactoryFactory.inputFactory()` instead of creating it on every call?

rrayst (changes requested):

core/src/main/java/com/predic8/membrane/core/interceptor/soap/SoapOperationExtractor.java
predic8: @rrayst 
Usually yes (see HardenedStaxInputFactory) . But this class is only used for the REST2SOAP converter that I think is obsoleted with the new template based rest to soap mapping in the tutorials and the book. Maybe we should take out REST2SOAP in version 8.

rrayst: Same code appears in `XMLElementFinder`.

rrayst (approved):

PR #3103, opened 2026-07-22
Add configurable PII masking for OpenAPI validation values (6.X downport)

predic8:
Downport of `feature/openapi-mask-validation-values` (commit c37ff34e4, targeting master/7.x) to the **6.X** line.

## What

Validation errors echo the submitted value (e.g. `"5 is smaller than the minimum of 18"`). A new opt-in `maskValues` attribute on `<openapi>` replaces that value with `***` so PII is not leaked.

Response and log are controlled independently:

```
maskValues: none | response | log | both   (default none, legacy behavior)
```

Masking is a render-time concern: the raw value stays in the `ValidationError` (held by a lazy `MaskableMessage` renderer) and is only masked when a message is rendered, so the response body and the server log can diverge (e.g. real value in the log for debugging, `***` in the response).

## Downport notes

Resolved 6.X-vs-7.x drift by keeping 6.X's validator structure and applying only the masking transformation:

- `ValidationContext` — kept 6.X (dropped a 7.x-only `content(Content)` helper not part of this feature).
- `StringRestrictionValidator` — masked 6.X's maxlen/minlen messages; did not pull in 7.x-only pattern/enum/const checks.
- `StringValidator` — kept 6.X's inline format validation; masked all value-echoing sites.
- `OpenAPIInterceptor` — added the `maskValues` param but kept 6.X's `router.isProduction()`.

## Testing

`MaskValuesTest` (4) and `OpenAPIRecordFactoryTest` (11) pass; `core` compiles.

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/openapi/serviceproxy/APIProxy.java
+++ b/core/src/main/java/com/predic8/membrane/core/openapi/serviceproxy/APIProxy.java
@@ -58,6 +58,8 @@ public class APIProxy extends ServiceProxy implements Polyglot, XMLSupport {
     public static final String RESPONSES = "responses";
     public static final String SECURITY = "security";
     public static final String VALIDATION_DETAILS = "details";
+    public static final String MASK_VALUES = "mask";
+    public static final String MASK_VALUES_LOG = "maskLog";
 
     private Language language = SPEL;
     private ExchangeExpression exchangeExpression;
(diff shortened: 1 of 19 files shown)


predic8:
@coderabbitai review

predic8:
I'll do a short review and come back

t-burch (approved):

Issue #3119, opened 2026-08-07
Incorrect OpenAPI-Validation with AllOf-Constructs


schiller-hank:
There seems to be an issue in the OpenAPI-validation with `allOf`-Constructs. The behviour was tested with the current version 7.4.0.
When using the following schema in a response of an OpenAPI specification
```
...
"Some-Object" : {
  "type" : "object",
  "properties" : {
    "type" : {
      "type" : "string",
      "enum" : [ "SOME-VALUE" ]
    }
  }
},
...
``` 
the validation in the gateway correctly responds
```
{
    "title": "OpenAPI message validation failed",
    "type": "https://membrane-api.io/problems/user/validation",
    "status": 500,
    "validation": {
        "method": "GET",
        "uriTemplate": "/some/path",
        "path": "/some/path",
        "schemaType": "string",
        "errors": {
            "RESPONSE/BODY#/type": [
                {
                    "message": "'ANOTHER-VALUE' is not part of the enum [SOME-VALUE].",
                    "complexType": "Some-Object",
                    "schemaType": "string"
                }
            ]
        }
    },
    "see": "https://membrane-api.io/problems/user/validation/openapi/response",
    "attention": "Membrane is running in development mode. Enable production mode in the configuration to prevent detailed information from appearing in error responses."
}
```
when the validated payload includes an invalid value (e.g. `ANOTHER-VALUE`) in the enum property.

However, when wrapping the above schema into an `allOf`-construct as in
```
...
"Some-Object" : {
  "type" : "object",
  "allOf" : [ {
    "type" : "object",
    "properties" : {
      "type" : {
        "type" : "string",
        "enum" : [ "SOME-VALUE" ]
      }
    }
  } ]
},
...
``` 
the gateway returns the json response payload with the invalid enum value although it is not valid as validated with the schema without `allOf` above.

predic8:
@schiller-hank thanks for reporting and the understandable description. Makes it easy for us to reproduce.
We'll have a look at the issue.

predic8:
@schiller-hank #3120 should fix this issue. The problem was a ref you probably used to. If you inline the ref your version should be fine.

We will release a bugfix version with the fix in the next couple of days.

schiller-hank:
@predic8 thank you very much for the quick reaction 🥇  Yes, there was a `$ref` used for that above mentioned schema. 

[...]

PR #3122, opened 2026-08-07
Lift the 3 MiB size limit on OpenAPI documents

predic8:
Reported by @schiller-hank in [#3119](https://github.com/membrane/api-gateway/issues/3119#issuecomment-3179): after inlining `$ref`s their OpenAPI document "hits a 3mb limit of the snakeyml or jackson library", which they could not configure, so they had to split the document.

## What the limit actually is

Not JVM memory — no amount of `-Xmx` affects it. It is SnakeYAML's `LoaderOptions.codePointLimit`, an **input size** guard, 3 MiB (`3145728`) by default, failing with `The incoming YAML document exceeds the limit: 3145728 code points`.

swagger-parser documents a `maxYamlCodePoints` system property for it. **That property could not fix Membrane**, because the limit is enforced in four places and the property reaches only one:

| # | Where | `-DmaxYamlCodePoints` reaches it? |
|---|---|---|
| 1 | `OpenAPIRecordFactory` — reads every document with a plain Jackson YAML mapper | **No** |
| 2 | swagger-parser SnakeYAML path (`DeserializationUtils.buildLoaderOptions()`) | Yes |
| 3 | swagger-parser `yaml30Mapper`/`yaml31Mapper` — built with **no** `LoaderOptions` | **No** — upstream [swagger-parser#2059](https://github.com/swagger-api/swagger-parser/issues/2059) |
| 4 | `OpenAPIUtil.convert2Json` — round-trips the parsed document when the API is published | **No** |

Leg 1 is decisive: `omYaml.readTree(...)` runs *before* swagger-parser on every path, so Membrane threw before the property could matter. Leg 3 is why anyone who *did* find the documented property still failed on 3.0/3.1 documents.

## The fix

There is no flag that switches the limit off — `codePointLimit` is an `int` that has to be set high. New `OpenAPIYamlLimits` sets all four legs to `Integer.MAX_VALUE` from one place, and still honours `maxYamlCodePoints` so an operator can put a bound back:

```sh
export JAVA_OPTS="-DmaxYamlCodePoints=10000000"
```

**This does not weaken billion-laughs protection** — that is a separate setting (`maxYamlAliasesForCollections`), left untouched. `codePointLimit` is only a size guard.

Legs 2 and 3 are global swagger-parser state, so they are set once from a static initialiser that runs before the first parse, with a comment explaining why.

## Tests

`OpenAPIYamlLimitsTest` generates an oversized document at runtime rather than committing a multi-megabyte fixture:

- `defaultMapperRejectsDocumentAboveThreeMebibytes` — pins the upstream default we work around, so we notice if swagger-parser changes it
- `readsDocumentAboveThreeMebibytes` — leg 1
- `parserReadsDocumentAboveThreeMebibytes` — legs 2 and 3, verified to fail with the exact reported error when the static initialiser is removed
- `codePointLimitIsLifted`

`870` tests pass in `com.predic8.membrane.core.openapi`.

## Note on the default

I defaulted to *unlimited* rather than merely making the knob work, on the grounds that specs are operator-supplied configuration and a cryptic startup failure is worse than a large parse. Happy to switch to a bounded default if you'd rather.

## Summary by CodeRabbit

* **New Features**
  * Added configurable YAML document size limits for OpenAPI processing.
  * OpenAPI YAML documents larger than the default limit can now be parsed successfully.

* **Bug Fixes**
  * Improved consistency between OpenAPI YAML parsing and Swagger parser handling for large documents.

* **Tests**
  * Added coverage verifying successful parsing of oversized, valid OpenAPI YAML documents and confirming the configured size limit is applied consistently.


core/src/main/java/com/predic8/membrane/core/openapi/util/OpenAPIYamlLimits.java
```diff
@@ -0,0 +1,102 @@
...
+public class OpenAPIYamlLimits {
+
+    private static final Logger log = LoggerFactory.getLogger(OpenAPIYamlLimits.class.getName());
+
+    /**
+     * Read by swagger-parser as well, so one setting covers Membrane and the parser. Read once when
+     * this class is loaded, so it has to be set at JVM startup:
+     * {@code JAVA_OPTS="-DmaxYamlCodePoints=..."}.
+     */
+    public static final String CODE_POINT_LIMIT_PROPERTY = "maxYamlCodePoints";
```
rrayst: Membrane's System Properties all start with `membrane.core.`

rrayst (changes requested):

rrayst (approved):

PR #3133, opened 2026-08-10
replace outdated scripts with scripts delegating to distribution/scripts

### LINKED ISSUE #3123: membrane.cmd ignores JAVA_OPTS, so JVM options cannot be set on Windows
## Problem

`distribution/router/membrane.cmd` — the Windows entry point shipped at the root of the distribution zip — invokes Java without `%JAVA_OPTS%`:

```bat
:: distribution/router/membrane.cmd:89-90
set "CLASSPATH=%membrane_home%\conf;%membrane_home%\lib\*"
java -cp "%CLASSPATH%" com.predic8.membrane.core.cli.RouterCLI %CMDARGS%
```

So on Windows there is no supported way to set heap, system properties, or GC flags. `export JAVA_OPTS=...` silently has no effect.

## Why it differs from Linux/macOS

The `.sh` launcher is a 20-line shim that delegates, and the delegate honours the variable:

```
membrane.sh → scripts/run-membrane.sh → scripts/java_check.sh → scripts/start_router.sh
```
```sh
# distribution/scripts/start_router.sh:88
java ${JAVA_OPTS:-} -cp "$CLASSPATH" com.predic8.membrane.core.cli.RouterCLI "$@"
```

`membrane.cmd` is instead **self-contained** (94 lines) and never delegates, so it misses the `JAVA_OPTS` handling that `scripts/start_router.cmd` already has:

```bat
:: distribution/scripts/start_router.cmd:47,49
set "JAVA_OPTS=%JAVA_OPTS% -Dmembrane.disable.term.colors=%DISABLE_COLORS% -Djdk.xml.maxGeneralEntitySizeLimit=0 -Djdk.xml.totalEntitySizeLimit=0"
java %JAVA_OPTS% -cp "%CLASSPATH%" com.predic8.membrane.core.cli.RouterCLI %*
```

Note the Windows path that *does* work (`scripts\run-membrane.cmd` → `start_router.cmd`) is not the one users are pointed at. Even one of our own examples, `distribution/examples/openapi/jwt-auth/membrane.cmd:90`, passes `%JAVA_OPTS%` — so the example is more correct than the shipped launcher.

Windows users also silently lose the two XML entity-limit properties that `start_router.cmd` sets, since those ride along in `JAVA_OPTS`.

## Impact

- `-Xmx`/`-Xms` cannot be set on Windows at all.
- `-Dlog4j.configurationFile=...` does not work, contradicting the documented usage in the `membrane.sh` header and `distribution/examples/logging/json/README.md`.
- `-DmaxYamlCodePoints=...` (see #3122) cannot be used to bound OpenAPI document size on Windows.
- `-Djdk.xml.*EntitySizeLimit` from `start_router.cmd` is not applied.

## Reproduce

```bat
set JAVA_OPTS=-Xmx123m
membrane.cmd
```
Then check the JVM: the flag is absent (`jcmd <pid> VM.flags`).

## Suggested fix

Make `membrane.cmd` a thin shim that delegates to `scripts\run-membrane.cmd`, mirroring what `membrane.sh` does, rather than duplicating a divergent Java invocation. That removes the duplication as well as the bug.

Found while investigating #3119, where the reporter's JVM-option settings appeared to be ignored.

christiangoerdes:
## Summary by CodeRabbit

- **Improvements**
  - Launchers now automatically locate the Membrane installation from nested directories.
  - Startup is handled consistently through the shared Membrane runner.
  - Command-line arguments and exit statuses are preserved.
  - Access-logging examples automatically apply their Log4j2 configuration.
  - Missing installations now produce a clear error.
  - OpenAPI JWT authentication examples continue supporting proxy configuration and custom Java options.


distribution/examples/logging/access/membrane.cmd
```diff
@@ -1,93 +1,26 @@
...
+set "SCRIPT_DIR=%~dp0"
+if "%SCRIPT_DIR:~-1%"=="\" set "SCRIPT_DIR=%SCRIPT_DIR:~0,-1%"
 
-where java >nul 2>&1
-if errorlevel 1 (
-    echo Java is not installed. Membrane needs at least Java %required_version%.
-    set "exitcode=1"
-    goto finish
-)
+set "JAVA_OPTS="-Dlog4j.configurationFile=%SCRIPT_DIR%\log4j2_access.xml" %JAVA_OPTS%"
```
rrayst: the quoting is wrong

rrayst (changes requested):

rrayst (approved):

Issue #3168, opened 2026-08-14
Flaky: Wsdl2OpenAPIXsdFeaturesTutorialTest races the mock backend on port 2001 under load


predic8:
`Wsdl2OpenAPIXsdFeaturesTutorialTest` fails intermittently when the machine is busy, with a varying subset of its 13 tests returning **504 Gateway Timeout**. On an idle machine it passes 13/13 reliably.

Observed failure counts across consecutive runs of one unchanged build: 8, 8, 3, 1, 3, 2 — a different subset each time. The same build then passed 13/13 three times in a row once the machine had settled.

## Symptom

```
INFO LogInterceptor {api=XSD Features} - ==== REQUEST ==== POST /features/get-simple-content
   ... 10s later ...
INFO HTTPClientInterceptor {api=XSD Features} - Target [http://localhost:2001/xsd-features] is not reachable.
INFO LogInterceptor {api=XSD Features} - HTTP/1.1 504 Gateway Timeout
```

The mock api never logs the request at all — the global `log` plugin in `mocks/xsd-features-service.apis.yaml` prints nothing for it — so the request is not being served on port 2001. The 10s gap is the HTTP client's socket timeout.

## Cause

`distribution/tutorials/soap/97-WSDL-XSD-Features.yaml` serves the gateway api on port **2000** and `include`s `mocks/xsd-features-service.apis.yaml`, whose mock apis listen on port **2001** — both in the same Membrane instance.

The test's readiness wait is satisfied before port 2001 is actually serving. In one captured run the failing request arrived **81 ms** after `Membrane API Gateway ... up and running!`, and it was the first request of the run:

```
08:11:00,335  INFO RouterCLI - Membrane API Gateway 7.5.1-SNAPSHOT up and running!
08:11:00,416  INFO LogInterceptor {api=XSD Features} - ==== REQUEST ==== POST /features/get-imported-type
08:11:10,438  INFO HTTPClientInterceptor - Target [http://localhost:2001/xsd-features] is not reachable.
```

Under load the window widens, which is why more tests fail the busier the machine is.

## Not a conversion bug

Ruled out `wsdl2openapi` itself: the generated OpenAPI document and the SOAP request bodies for all 14 operations of `mocks/xsd-features.wsdl` were byte-identical before and after the code change originally suspected. The failures reproduce with unmodified `core` as well, given enough load.

## Suggested fix

[...]

PR #3170, opened 2026-08-14
fix: retry a request when establishing the connection timed out

predic8:
## Problem

A timeout while *connecting* is transient and safe to retry — no byte of the request has been sent, so the target cannot have processed it. Membrane gave up on it instead.

The reason is that the JDK reports both phases as a bare `java.net.SocketTimeoutException`, differing only in message text ("Connect timed out" vs "Read timed out"). `RetryHandler` could therefore only apply the conservative read-timeout rule — retry only idempotent methods, and only when several nodes are configured — so a `POST` to a single backend was never retried.

This is the root cause of the flaky `Wsdl2OpenAPIXsdFeaturesTutorialTest` in #3168. A gateway self-calling a mock on a fixed loopback port eventually draws an ephemeral port whose 4-tuple against that port is still in `TIME_WAIT`; the kernel silently drops the SYN, the connect times out after the full 10s, and the backend never sees the connection at all. Captured during a hang:

```
127.0.0.1.62506 -> 127.0.0.1.2001   SYN_SENT     <- the new outbound connect
127.0.0.1.2001  -> 127.0.0.1.62506  TIME_WAIT    <- an earlier connection, same 4-tuple
*.2001                              LISTEN  Recv-Q=0
"Connection Acceptor '*:2001'"      RUNNABLE at sun.nio.ch.Net.accept()
```

The readiness race suggested in the issue and a full accept backlog were both ruled out: `Recv-Q` is 0, the acceptor thread is healthy, and both ports are bound before `up and running!` is logged.

## Change

- **`ConnectTimeoutException extends SocketTimeoutException`** — thrown by `Connection.open` for everything that happens before the request is written: the plain `connect`, both TLS `createSocket` paths, and the post-tunnel handshake. Extending `SocketTimeoutException` keeps every existing catch site working.
- **`RetryHandler`** retries it for *any* request method, controlled by a new attribute, default `true`:
  ```yaml
  retries:
    retryOnConnectTimeout: false   # fail fast instead
  ```
- **Read timeouts keep the existing rule.** There the backend may already have processed the request, so retrying a non-idempotent method could duplicate a side effect.
- **`HTTPClientInterceptor` now names the phase.** Previously a connect timeout, a read timeout and a refused connection all logged `Target ... is not reachable.`, which is what made this issue misdiagnose:
  ```
  Target http://127.0.0.1:2999/ timed out while the connection was being established, no request was sent. Reason: Connecting to 127.0.0.1:2999 timed out after 10000ms.
  Target http://localhost:62910/ timed out while waiting for the response, the request had already been sent. Reason: Read timed out
  ```
  `subSee` is now `connect-timeout` vs `socket-timeout`, and the 504 problem details carry a `detail` (they had none).

## Verification

- `ConnectionTest.connectTimeoutSurfacesAsConnectTimeoutException` is a real regression test — with the `Connection` change stashed it fails with *expected `ConnectTimeoutException` but was `java.net.SocketTimeoutException`*.
- `RetryHandlerTest` covers connect timeout retried for `POST` on one node, not retried when the flag is off, and read timeouts unchanged.
- End-to-end against a backend whose accept queue is full: **30.4s = 3 attempts × 10s**, with `Attempt #0/#1/#2` and *"timed out before it was established"* in the log. Before the change: one attempt, immediate 504.

## Notes

- An unlucky call is now slower rather than failed — up to 3 × `connectTimeout` (30s at defaults) before the 504. `retryOnConnectTimeout: false` restores fail-fast.
- The tutorial and IT from #3168 live on a feature branch, not on master, so the flake itself cannot be exercised here. Confirming it requires running that IT (~1 in 5 runs failed at baseline) once this is merged into that branch.
- Not addressed here: the tutorial ITs start Membrane in `@BeforeEach`, i.e. once per test method, which is what accumulates the `TIME_WAIT` entries in the first place. Worth a separate issue.

Refs #3168

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timeout handling by distinguishing connection-establishment failures from response-read timeouts.
  * HTTP error responses now include clearer status codes and diagnostic details for connection and response delays.
  * Connection timeout errors include the affected endpoint and timeout information.

* **New Features**
  * Connection-establishment timeouts can be retried for all request methods by default.
  * Added configuration to enable or disable connection-timeout retries.
  * Read-timeout retry behavior remains governed by request safety and destination availability.

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/interceptor/HTTPClientInterceptor.java
+++ b/core/src/main/java/com/predic8/membrane/core/interceptor/HTTPClientInterceptor.java
@@ -19,6 +19,7 @@ import com.predic8.membrane.annot.MCElement;
 import com.predic8.membrane.core.exchange.Exchange;
 import com.predic8.membrane.core.http.MalformedHeaderException;
 import com.predic8.membrane.core.proxies.AbstractServiceProxy;
+import com.predic8.membrane.core.transport.http.ConnectTimeoutException;
 import com.predic8.membrane.core.transport.http.EOFWhileReadingLineException;
 import com.predic8.membrane.core.transport.http.HttpClient;
 import com.predic8.membrane.core.transport.http.ProtocolUpgradeDeniedException;
@@ -119,12 +120,21 @@ public class HTTPClientInterceptor extends AbstractInterceptor {
                     .buildAndSetResponse(exc);
             return ABORT;
         } catch (SocketTimeoutException e) {
-            // Details are logged further down in the HTTPClient
-            log.info("Target {} is not reachable.",exc.getDestinations());
+            // Name the phase: a connect timeout means the target never accepted the connection, a read
+            // timeout means it accepted but did not answer. Both used to be logged as
+            // "is not reachable.", which is also what a refused connection logs, so the three were
+            // indistinguishable in the log.
+            boolean whileConnecting = e instanceof ConnectTimeoutException;
+            var msg = "Target %s timed out %s.".formatted(getDestination(exc),
+                    whileConnecting ? "while the connection was being established, no request was sent"
+                                    : "while waiting for the response, the request had already been sent");
+            log.info("{} Reason: {}", msg, e.getMessage());
             internal(router.getConfiguration().isProduction(), getDisplayName())
                     .title("Gateway Timeout")
                     .status(504)
-                    .addSubSee("socket-timeout")
+                    .addSubSee(whileConnecting ? "connect-timeout" : "socket-timeout")
+                    .detail(msg)
+                    .stacktrace(false)
                     .buildAndSetResponse(exc);
             return ABORT;
         } catch (UnknownHostException e) {
--- a/core/src/main/java/com/predic8/membrane/core/proxies/RuleManager.java
+++ b/core/src/main/java/com/predic8/membrane/core/proxies/RuleManager.java
@@ -14,22 +14,28 @@
 
 package com.predic8.membrane.core.proxies;
 
-import com.predic8.membrane.core.exchange.*;
-import com.predic8.membrane.core.exchangestore.*;
-import com.predic8.membrane.core.http.*;
-import com.predic8.membrane.core.model.*;
-import com.predic8.membrane.core.router.*;
-import com.predic8.membrane.core.transport.http.*;
-import com.predic8.membrane.core.transport.ssl.*;
-import com.predic8.membrane.core.util.*;
-import org.jetbrains.annotations.*;
-import org.slf4j.*;
+import com.predic8.membrane.core.exchange.Exchange;
+import com.predic8.membrane.core.exchangestore.ExchangeStore;
+import com.predic8.membrane.core.http.Request;
+import com.predic8.membrane.core.router.Router;
+import com.predic8.membrane.core.transport.http.AbstractHttpHandler;
+import com.predic8.membrane.core.transport.http.IpPort;
+import com.predic8.membrane.core.transport.ssl.SSLContext;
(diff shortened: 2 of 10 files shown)


predic8:
Thanks — went through all four. Two fixed in 23b5602, two I'm pushing back on with evidence.

### 1. Close the partially opened socket — ✅ fixed, but the rationale was only half right

Fixed defensively, with the close failure attached as `suppressed`.

Worth recording that the stated failure mode does **not** apply to the plain-connect path: `Socket.connect` closes the socket itself on timeout, so nothing leaked there. Measured over 40 timed-out connects against a filled accept queue, counting the process's own descriptors:

```
with the fix:    fds before=201 after=204 delta=3
without the fix: fds before=201 after=204 delta=3
```

Identical. What *is* real is the case named in the finding's second half: if `doTunnelHandshake` or the TLS wrapping times out, `con.socket` is by then a fully connected socket and was discarded unclosed. That is what the fix covers, and it matters more now that a connect timeout is retried — the leak would be per attempt rather than one-off.

### 2. Do not expose the backend destination in the 504 detail — ❌ declining

Two reasons.

**Production mode already prevents this.** `ProblemDetails` drops `detail` entirely when production is on, substituting a log key (`ProblemDetails.java:247-250`):

```java
if (production) {
    provideLogKeyInsteadOfDetails(root);
    return root;
}
if (detail != null) {
    root.put(DETAIL, detail);
}
```

In development mode the response even carries an explicit banner saying detailed information is being exposed and how to turn it off. So the destination reaches a client only where that is the deliberate, signposted behaviour.

**It would be inconsistent with the same method.** The neighbouring branches on master already put the destination in `detail` — `ConnectException` (`"Target %s is not reachable."` with `getDestination(exc)`), `SocketException` (`exc.getDestinations()`), `UnknownHostException` (the target authority). Stripping it from only the timeout branch would make 504 less informative than 502 for no security gain, since the exposure is governed by the production flag rather than by the individual branch.

Happy to revisit if the project wants destinations out of `detail` everywhere, but that is a separate, cross-cutting change.

### 3. `{@code}` → HTML markup — ❌ declining

The guideline applies to the doc tags the custom generator renders to membrane-api.io (`@description`, `@example`, ...). My new `@description` on `setRetryOnConnectTimeout` does use `<code>true</code>` / `<code>false</code>` accordingly.

The line flagged here is class-level Javadoc **prose outside any of those tags**, which the generator drops rather than renders, so it never reaches the published reference. It is ordinary IDE Javadoc, and the immediately preceding list item — pre-existing — reads `{@code failOverOn5XX=true}`. Changing only my line would leave two adjacent items in different styles.

### 4. Add a handler-level connect-timeout test — ✅ fixed, good catch

That branch really was untested; I had only verified it end-to-end by hand. Added `connectTimeoutYields504NamingTheConnectPhase`, injecting an `HttpClient` that throws `ConnectTimeoutException` and asserting 504 plus the `connect-timeout` subtype. `HTTPClientInterceptorTest` is now 16/16, `RetryHandlerTest` 19/19.

---

One note for human reviewers: `ConnectionTest` cannot be run on my machine right now. My own load testing saturated the loopback `TIME_WAIT` table (1610 entries on port 2000, non-expiring), and that class binds fixed port 2000 and self-connects in `setUp` — so it hits the very bug this PR is about. I verified `Connection.open`'s behaviour out-of-band instead: `ConnectTimeoutException: Connecting to 127.0.0.1:52875 timed out after 200ms.` with the plain `SocketTimeoutException` as cause. CI should run it cleanly.

predic8:
@coderabbitai review

rrayst (approved):

PR #3186, opened 2026-08-31
Fix/sticky body read failure

predic8:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of interrupted or incomplete request and response bodies.
  * Failed body reads now consistently report the original error and prevent unsafe repeated stream access.
  * Added clearer detection and logging for client disconnects and upstream body failures.
  * Observers are notified reliably when body reading fails, including when registered afterward.
* **Tests**
  * Added coverage for interrupted streams, chunked-body failures, repeated access, observer notifications, and disconnect detection.

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/http/AbstractBody.java
+++ b/core/src/main/java/com/predic8/membrane/core/http/AbstractBody.java
@@ -14,12 +14,15 @@
 
 package com.predic8.membrane.core.http;
 
-import org.slf4j.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import java.io.*;
-import java.util.*;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.List;
 
-import static java.nio.charset.StandardCharsets.*;
+import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
  * A HTTP message body (request or response), as it is received or constructed
@@ -43,6 +46,12 @@ import static java.nio.charset.StandardCharsets.*;
  * threads is illegal. Using a Body Stream after the Body as been accessed by
  * someone else (using streams or not) is illegal.
  * <p>
+ * Failing to read a body is terminal: the {@link ReadingBodyException} is recorded (see
+ * {@link #setObservedException(ReadingBodyException)}), the observers are notified via
+ * {@link MessageObserver#bodyError(ReadingBodyException)} and every later access re-throws that same
+ * exception instead of touching the (dead) stream again. This keeps the original cause visible rather
+ * than replacing it with a follow-up error like "stream is closed".
+ * <p>
  * Public instance methods must not throw {@link IOException}s. Throw an
  * unchecked {@link ReadingBodyException} or {@link WritingBodyException} instead.
  * (This is enforced by the BodyDoesntThrowIOExceptionTest .)
@@ -63,6 +72,8 @@ public abstract class AbstractBody {
 		if (read)
 			return;
 
+		throwIfFailed();
+
 		if (wasStreamed)
 			throw new IllegalStateException("Cannot read body after it was streamed.");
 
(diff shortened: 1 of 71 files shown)


predic8:
@coderabbitai review

core/src/main/java/com/predic8/membrane/core/http/AbstractBody.java
```diff
@@ -78,17 +89,79 @@ public void read() {
...
+	 * <p>
+	 * The first failure wins: the recorded exception is never replaced and the observers are notified
+	 * only once. The returned exception is the one the caller passed in, so that the idiom
+	 * <code>throw setObservedException(new ReadingBodyException(e));</code> keeps throwing the exception
+	 * belonging to the current failure.
+	 */
     protected ReadingBodyException setObservedException(ReadingBodyException e) {
-        return observedException = e;
+        if (observedException != null)
+            return e;
```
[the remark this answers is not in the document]
predic8: @coderabbitai

Verification against current code

 - Every public entry point already latches before touching the stream: read(), getContent(), getContentAsStream(), write(), getRaw() (via read()) call throwIfFailed(), which throws the recorded instance; discard() and ChunkedBody.read()/write()/discard() check observedException explicitly. So setObservedException() is only ever reached on a first failure — its early-return branch is defensive, not a live path, and cannot leak a different instance to a caller.
 - The behaviour is deliberate and documented in the method's Javadoc: the returned exception is the caller's so the throw setObservedException(new ReadingBodyException(e)) idiom throws the exception belonging to the current failure. ChunkedBody.getContentAsStream() (line ~216) relies on calling it without throwing the return value.
 - The requested regression test already exists and already passes:
   - core/src/test/java/com/predic8/membrane/core/http/ChunkedBodyStickyErrorTest.java:54 readAfterFailureDoesNotDrainTheDeadStream — lazy-stream failure, then body.read(), assertSame(e, ...).
   - core/src/test/java/com/predic8/membrane/core/http/BodyStickyErrorTest.java:53-77 — same-instance assertions across read(), all accessors, and getObservedException().

core/src/main/java/com/predic8/membrane/core/util/ExceptionUtil.java
```diff
@@ -43,6 +47,20 @@ public static String concatMessageAndCauseMessages(Throwable throwable) {
...
+     * upload. Such events are normal operation and should not be logged as errors.
+     * <p>
+     * Deliberately type-based: a plain {@link java.io.IOException} is not treated as a disconnect,
+     * because it can just as well indicate a genuine fault.
+     */
+    public static boolean isPeerDisconnect(Throwable t) {
+        Throwable root = getRootCause(t);
+        return root instanceof ClosedChannelException
+                || root instanceof SocketException
+                || root instanceof EOFException;
```
rrayst, line 75: This also treats `ConnectException` as a "peer disconnect".

predic8: Thanks, good catch — fixed in 81206db.

`isPeerDisconnect()` now returns `false` up front for the `SocketException` subclasses that mean "the connection could never be established" rather than "the peer went away": `ConnectException`, `NoRouteToHostException`, `PortUnreachableException` and `BindException`. The broad `SocketException` check stays, so "Connection reset"/"Broken pipe" are still matched without resorting to message matching. Javadoc explains the exclusions; four tests added in `ExceptionUtilTest`.

For the record it was not an active bug: the only call sites (`HttpServletHandler:94` and `logReadingBodyException()`) only catch `ReadingBodyException`/`WritingBodyException`, i.e. the connection is already established there, so a `ConnectException` cannot reach them. This fixes the contract of the method rather than observed behaviour. Also note `SocketTimeoutException` was never affected — it extends `InterruptedIOException`, not `SocketException` — but an OS-level connect timeout surfaces as `ConnectException: Operation timed out` and is covered now.

predic8:
Follow-up from the review discussion: while checking whether a failed body read can leave a connection reusable, two pre-existing bugs turned up — the client connection is kept alive although it is desynchronized, and the target connection is orphaned when the request body fails while being streamed to the backend. Written up in #3187.

Both predate this PR, but the fix builds on the `hasFailed()` flag introduced here, so it rides along in this branch: `Exchange.canKeepConnectionAlive()` now also consults the body state, `HttpServerHandler` closes the orphaned target connection in its `finally`, plus `ExchangeKeepAliveTest`.

rrayst (approved):

PR #3189, opened 2026-09-02
fix(openapi): honour nullable when allOf/oneOf/anyOf is present (#3119)

predic8:
The null/nullable short-circuit in SchemaValidator ran after the composition keywords had already been validated, so a `nullable: true` schema carrying an `allOf`/`oneOf`/`anyOf`/`not` descended into subschemas that are typically not nullable and rejected `null` with "Value null is not an object." Without a composition keyword the short-circuit was reached first, which is why inlining the referenced properties made the same `null` pass.

The check now sits directly after $ref resolution, before the composition dispatch. isNullable() is unchanged, so OAS 3.1 `type: [..., "null"]` is covered as well. Request and response bodies share SchemaValidator.

Reported by @schiller-hank in #3119.

## Summary by CodeRabbit

* **Bug Fixes**
  * Nullable values now validate correctly with `allOf`, `oneOf`, `anyOf`, and `not` schemas.
  * Non-nullable schemas continue to reject null values.
  * Object values within nullable composed schemas are still fully validated.
  * Nullable request-body scenarios now behave consistently across composed schemas.

* **Tests**
  * Added coverage for nullable composition scenarios, non-nullable validation, and nested object validation.

### DIFF
--- a/core/src/main/java/com/predic8/membrane/core/openapi/validators/SchemaValidator.java
+++ b/core/src/main/java/com/predic8/membrane/core/openapi/validators/SchemaValidator.java
@@ -110,6 +110,11 @@ public class SchemaValidator implements JsonSchemaValidator {
                 throw new RuntimeException("Should not happen!");
         }
 
+        // A nullable schema accepts null itself, so validation stops here instead of descending into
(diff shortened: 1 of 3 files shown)


predic8:
@coderabbitai review

rrayst (approved):

PR #3204, opened 2026-09-04
Check response flow in dynamic registration

### LINKED ISSUE #3200: DynamicRegistration ignores the response-flow outcome and returns a possibly stale response
`DynamicRegistration.doRequest` ignores the response-flow outcome and returns a possibly stale response.

## Where

`core/src/main/java/com/predic8/membrane/core/interceptor/oauth2/authorizationservice/DynamicRegistration.java`, `doRequest(Exchange exc)` (lines 73–88):

```java
if (router.getFlowController().invokeRequestHandlers(exc, interceptors) != CONTINUE)
    throw new RuntimeException("Registration interceptorchain (request) had a problem");

client.call(exc);
Response response = exc.getResponse();

router.getFlowController().invokeResponseHandlers(exc, interceptors);

if (response.getStatusCode() < 200 || response.getStatusCode() > 201)
    throw new RuntimeException("Registration endpoint didn't return successful: " + response.getStatusMessage());
return response;
```

## Two problems

**1. The response flow's outcome is discarded.** The request flow directly above it throws on anything but `CONTINUE`; the response flow's return value is dropped on the floor. `invokeResponseHandlers` returns `ABORT` when one of the configured interceptors aborts the response flow or throws in `handleResponse`, and that failure is silently ignored here — registration then continues on a response that an interceptor considered unusable.

**2. `response` is captured before the response handlers run.** Any interceptor that *replaces* the response on the exchange (rather than mutating it in place) leaves the local `response` stale. The status check on the next line and the value handed back to the caller are then the pre-interceptor response — so, for instance, a response the flow rewrote or replaced with an error is not what `registerWithCallbackAt` / `retrieveOpenIDConfiguration` end up parsing.

## Suggested direction

Mirror the request-flow guard and read the response after the handlers have run:

```java
client.call(exc);

if (router.getFlowController().invokeResponseHandlers(exc, interceptors) != CONTINUE)
    throw new RuntimeException("Registration interceptorchain (response) had a problem");

Response response = exc.getResponse();
```

## Notes

Reached in production via `MembraneAuthorizationService` (`<membrane>` authorization service with a `<registration>` child) for OpenID configuration discovery and dynamic client registration, both at startup. There are no tests for this class and no `<registration>` usage in the repo's configs, tutorials, or examples, so a fix should come with coverage.

Pre-existing; found while auditing `FlowController` callers.

utafrali:
Hit an issue where response handlers in dynamic registration weren't being checked, so we'd grab stale responses. Moved the read to after the handlers run and added a check for the flow outcome. Included tests. Fixes #3200.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved OAuth2 dynamic registration response handling.
  - Responses interrupted by processing rules now fail correctly instead of being silently ignored.
  - Responses replaced during processing are now returned as expected.

- **Tests**
  - Added coverage for successful configuration retrieval, interrupted responses, and responses replaced during processing.

### DIFF
--- a/core/src/test/java/com/predic8/membrane/core/interceptor/oauth2/authorizationservice/DynamicRegistrationTest.java
+++ b/core/src/test/java/com/predic8/membrane/core/interceptor/oauth2/authorizationservice/DynamicRegistrationTest.java
@@ -1,0 +1,94 @@
+/* Copyright 2025 predic8 GmbH, www.predic8.com
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License. */
+
+package com.predic8.membrane.core.interceptor.oauth2.authorizationservice;
+
+import com.predic8.membrane.core.exchange.*;
+import com.predic8.membrane.core.http.*;
+import com.predic8.membrane.core.interceptor.*;
+import com.predic8.membrane.core.router.*;
+import com.predic8.membrane.core.transport.http.*;
+import com.predic8.membrane.core.transport.http.client.*;
+import org.junit.jupiter.api.*;
+
+import java.io.*;
+
+import static com.predic8.membrane.core.interceptor.Outcome.*;
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.ArgumentMatchers.*;
+import static org.mockito.Mockito.*;
+
+class DynamicRegistrationTest {
+
+    private DynamicRegistration reg;
+    private FlowController flowController;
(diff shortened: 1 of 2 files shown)


utafrali:
Got it, I'll fix the response cleanup on abort and improve the replacement test coverage.

core/src/test/java/com/predic8/membrane/core/interceptor/oauth2/authorizationservice/DynamicRegistrationTest.java
```diff
@@ -0,0 +1,94 @@
+/* Copyright 2025 predic8 GmbH, www.predic8.com
```
predic8, line 1: 2025 => 2026

predic8 (changes requested):

utafrali:
done, updated the year

predic8:
@utafrali did you push?
Cloud u also address the comments from @coderabbitai ?

utafrali:
On it, pushing the fixes now.

lsiepel:
Be aware, this bot (PR author) is creating many PR’s in different repo’s. Not understanding comments given. Adding untested code, not understanding the issue.
Many are closed after maintainers have spend some time on it, no human interaction at all.

predic8:
@lsiepel thanks for saving me the time

PR #3213, opened 2026-09-07
Trigger Docker image build explicitly after a release

rrayst:
## Problem

`Build Docker Image` only triggers on `release: [published]`. Since #3085 switched the release creation in `release-build.yml` from `secrets.PAT` to `${{ github.token }}`, that event no longer starts a workflow: GitHub [suppresses workflow triggering for events created with `GITHUB_TOKEN`](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow).

Run history confirms it — the last `release`-triggered Docker build was **v7.2.5 (2026-06-26)**; everything since was manual. **v6.6.0 and v7.4.0 have no image on Docker Hub.**

## Changes

**`release-build.yml`**
- Adds an explicit `gh workflow run docker-publish.yml` step after the release is created. `workflow_dispatch` and `repository_dispatch` are the documented exceptions to the `GITHUB_TOKEN` restriction, so this works with the default token — it just needs `actions: write`.
- Dispatched on `github.base_ref` (the release branch) with the tag passed as an input, to avoid the branch/tag name ambiguity of `v<version>`.

**`docker-publish.yml`**
- New `tag` input: the release tag to build. Falls back to `github.ref_name` when the workflow is run on the tag itself, so dispatching on a tag keeps working. Validated, with a clear error if it is not a release tag.
- New `latest` input (default `false`) and a corrected `latest` condition. Previously `enable=${{ ! github.event.release.prerelease }}` evaluated to **true** on any manual run, because `github.event.release` is empty there — that is how the manual v7.5.0 run came to repoint `predic8/membrane:latest`. Now `latest` follows the release for a `release` event and the input for a manual run.
- Checkout and the release download both use the resolved tag (`refs/tags/...`), so a build dispatched from a branch still builds the right release.
- Drops the `bhowell2/github-substring-action` dependency; the version is derived in the resolve step.

## Notes

- `release-build.yml` hardcodes `prerelease: true`, so a release-triggered run still never pushes `latest` — unchanged behaviour. Pushing `latest` stays a deliberate manual dispatch with `latest=true`.
- **master needs the same change** — it has the identical `github.token` + `release` trigger combination.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -52,7 +52,7 @@ jobs:
         run: |
           set -euo pipefail
           RELEASE_TAG="${INPUT_TAG:-$REF_NAME}"
-          if [[ ! "$RELEASE_TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+ ]]; then
+          if [[ ! "$RELEASE_TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?([+][0-9A-Za-z.-]+)?$ ]]; then
             echo "::error::'$RELEASE_TAG' is not a release tag. Dispatch this workflow on the tag, or pass the 'tag' input (e.g. v6.6.0)."
             exit 1
           fi
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -96,11 +96,16 @@ jobs:
       # 'latest' follows the release for a real release, and the 'latest' input
       # for a manual run. Without the explicit event check a manual run would
       # always push 'latest', because github.event.release is empty there.
+      # 'flavor: latest=false' is required as well: metadata-action defaults to
+      # 'latest=auto', which adds 'latest' for any semver version regardless of
+      # the enable= condition on the raw tag below.
       - name: Extract Docker metadata
         id: meta
         uses: docker/metadata-action@v4
         with:
           images: "${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}"
+          flavor: |
+            latest=false
           tags: |
             type=semver,pattern={{version}},value=${{ steps.release_version.outputs.release_tag }}
             type=semver,pattern={{major}}.{{minor}},value=${{ steps.release_version.outputs.release_tag }}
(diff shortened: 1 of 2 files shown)


rrayst:
### Follow-up: `flavor: latest=false`

The manual v6.6.0 build (run 34100932880) pushed `6.6.0`, `6.6`, `6` — and `latest`, despite `type=raw,value=latest,enable=false`.

Cause: `docker/metadata-action` defaults to `flavor: latest=auto`, which appends `latest` for any semver version **independently** of the `enable=` condition on the raw entry. So that condition has been dead code all along — every release build pushed `latest`, including maintenance releases from older branches (e.g. the v6.5.3 run on 2026-06-26 pushed `latest` too; it only survived because the v7.2.5 build ran ten minutes later).

Added `flavor: latest=false` so the explicit `type=raw,value=latest,enable=...` entry is the only source of `latest`.

⚠️ `predic8/membrane:latest` currently points at **6.6.0** and needs to be repointed at 7.5.0.

rrayst:
Carried over from the review on #3214: the release tag check is now anchored to a complete SemVer version.

The unanchored pattern accepted values like `v7.4.0junk`, for which `metadata-action` emits no semver tags — so a run with `latest=true` would have published the image as `latest` only. A bare `$` would have rejected `v7.0.0-RC1`, which is a real tag here, so the optional prerelease and build-metadata parts are matched as well:

```bash
^v[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?([+][0-9A-Za-z.-]+)?$
```

Accepts `v6.6.0`, `v7.4.0`, `v7.0.0-RC1`, `v6.3.15`, `v7.4.0+build.5`; rejects `v7.4.0junk`, `v7.4.0.1`, `v7.4`, `master`, `6.X`.

Also worth carrying over: the v6.6.0 dispatch (run 34100932880) ran **before** the `flavor: latest=false` commit and did push `latest`, which was repointed to 7.5.0 by hand afterwards. So that fix is still unexercised — check the metadata step's `Processing flavor input` / `Docker tags` groups on the next release.

predic8 (approved):

PR #3214, opened 2026-09-07
Trigger Docker image build explicitly after a release (port of #3213)

rrayst:
Port of #3213 to `master`. Same root cause, same fix — `master` carries the identical `github.token` + `release`-trigger combination, which is why **v7.4.0 has no image on Docker Hub** either.

## Problem

`Build Docker Image` only triggers on `release: [published]`. Since #3085 switched the release creation in `release-build.yml` from `secrets.PAT` to `${{ github.token }}`, that event no longer starts a workflow: GitHub [suppresses workflow triggering for events created with `GITHUB_TOKEN`](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow).

The last `release`-triggered Docker build was **v7.2.5 (2026-06-26)**; v7.3.1 and v7.5.0 were built by manual dispatch, and v7.4.0 was never built at all.

## Changes

**`release-build.yml`**
- Adds an explicit `gh workflow run docker-publish.yml` step after the release is created. `workflow_dispatch` and `repository_dispatch` are the documented exceptions to the `GITHUB_TOKEN` restriction, so this works with the default token — it just needs `actions: write`.
- Dispatched on `github.base_ref` with the tag passed as an input, to avoid the branch/tag name ambiguity of `v<version>` (`release-pr.yml` creates both).

**`docker-publish.yml`**
- New `tag` input, falling back to `github.ref_name` so dispatching on a tag keeps working; validated, with a clear error otherwise. Checkout and the release download both use the resolved `refs/tags/<tag>`, so a build dispatched from a branch still builds the right release.
- New `latest` input (default `false`) plus two fixes to the `latest` tag:
  - `flavor: latest=false`. `docker/metadata-action` defaults to `latest=auto`, which appends `latest` for **any** semver version independently of the `enable=` condition on the raw entry — so that condition was dead code, and every release build pushed `latest`, including maintenance releases from older branches.
  - `enable=${{ ! github.event.release.prerelease }}` evaluates to **true** on a manual run, because `github.event.release` is empty there. Now `latest` follows the release for a `release` event and the input for a manual run.
- Drops the `bhowell2/github-substring-action` dependency; the version is derived in the resolve step.

## Notes

- `release-build.yml` hardcodes `prerelease: true`, so a release-triggered build never publishes `latest`. Pushing `latest` is a deliberate dispatch with `latest=true` — which is what already happened for v7.5.0.
- Merge #3213 first, or at least don't merge this alone and then cut a release from `6.X`: the dispatch step needs `docker-publish.yml` with the `tag` input present on the branch it dispatches to.
- v6.6.0 has meanwhile been built and pushed manually ([run 34100932880](https://github.com/membrane/api-gateway/actions/runs/34100932880)). **v7.4.0 has not** — decide whether it is still worth publishing.

## Summary by CodeRabbit

- **New Features**
  - Manual Docker publishing supports specifying a release tag.
  - Manual publishing can optionally apply the `latest` image tag.
  - Release builds automatically trigger the corresponding Docker image build.

- **Bug Fixes**
  - Docker image builds and publishing now follow the release workflow reliably.
  - Image tags and release downloads consistently use the selected release version.
  - Prerelease versions no longer receive the `latest` tag automatically.
  - Release tags are validated before Docker publishing.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -52,7 +52,7 @@ jobs:
         run: |
           set -euo pipefail
           RELEASE_TAG="${INPUT_TAG:-$REF_NAME}"
-          if [[ ! "$RELEASE_TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+ ]]; then
+          if [[ ! "$RELEASE_TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?([+][0-9A-Za-z.-]+)?$ ]]; then
             echo "::error::'$RELEASE_TAG' is not a release tag. Dispatch this workflow on the tag, or pass the 'tag' input (e.g. v7.4.0)."
             exit 1
           fi
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -9,7 +9,18 @@ on:
   release:
     types: [published]
   workflow_dispatch:
-    
+    inputs:
+      tag:
+        description: "Release tag to build, e.g. v7.4.0. Required unless the workflow is run on the tag itself."
+        type: string
+        required: false
+        default: ''
+      latest:
+        description: "Also push the 'latest' Docker tag"
+        type: boolean
+        required: false
+        default: false
+
 env:
   REGISTRY: docker.io
   IMAGE_NAME: predic8/membrane
@@ -29,16 +40,30 @@ jobs:
     steps:
 #      - name: Dump context
 #        uses: crazy-max/ghaction-dump-context@v2
-      - uses: bhowell2/github-substring-action@1.0.2
+
+      # The release tag can come from the 'tag' input (when dispatched from a
+      # branch, e.g. by the Create Release workflow) or from the ref the
+      # workflow runs on (when triggered by a release, or dispatched on a tag).
+      - name: Resolve release tag
         id: release_version
-        with:
-          value: "${{ github.ref_name }}"
-          index_of_str: "v"
-          output_name: tag
+        env:
+          INPUT_TAG: ${{ inputs.tag }}
+          REF_NAME: ${{ github.ref_name }}
+        run: |
+          set -euo pipefail
+          RELEASE_TAG="${INPUT_TAG:-$REF_NAME}"
+          if [[ ! "$RELEASE_TAG" =~ ^v[0-9]+[.][0-9]+[.][0-9]+ ]]; then
+            echo "::error::'$RELEASE_TAG' is not a release tag. Dispatch this workflow on the tag, or pass the 'tag' input (e.g. v7.4.0)."
+            exit 1
+          fi
+          echo "release_tag=$RELEASE_TAG" >> "$GITHUB_OUTPUT"
+          echo "tag=${RELEASE_TAG#v}" >> "$GITHUB_OUTPUT"
+          echo "Building Docker image for release $RELEASE_TAG"
+
       - name: Checkout repository
         uses: actions/checkout@v7
         with:
-          ref: ${{ github.ref }}
+          ref: refs/tags/${{ steps.release_version.outputs.release_tag }}
           persist-credentials: false
 
       # Install the cosign tool except on PR
@@ -69,21 +94,29 @@ jobs:
 
       # Extract metadata (tags, labels) for Docker
       # https://github.com/docker/metadata-action
+      # 'latest' follows the release for a real release, and the 'latest' input
+      # for a manual run. Without the explicit event check a manual run would
+      # always push 'latest', because github.event.release is empty there.
+      # 'flavor: latest=false' is required as well: metadata-action defaults to
+      # 'latest=auto', which adds 'latest' for any semver version regardless of
+      # the enable= condition on the raw tag below.
       - name: Extract Docker metadata
         id: meta
         uses: docker/metadata-action@v4
         with:
           images: "${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}"
+          flavor: |
+            latest=false
           tags: |
-            type=semver,pattern={{version}}
-            type=semver,pattern={{major}}.{{minor}}
-            type=semver,pattern={{major}}
-            type=raw,value=latest,enable=${{ ! github.event.release.prerelease }}
+            type=semver,pattern={{version}},value=${{ steps.release_version.outputs.release_tag }}
+            type=semver,pattern={{major}}.{{minor}},value=${{ steps.release_version.outputs.release_tag }}
+            type=semver,pattern={{major}},value=${{ steps.release_version.outputs.release_tag }}
+            type=raw,value=latest,enable=${{ (github.event_name == 'workflow_dispatch' && inputs.latest) || (github.event_name == 'release' && ! github.event.release.prerelease) }}
             type=sha,enable=false
 
       - uses: robinraju/release-downloader@v1.8
         with:
-          tag: "${{ github.ref_name }}"
+          tag: "${{ steps.release_version.outputs.release_tag }}"
           fileName: "membrane-api-gateway-${{ steps.release_version.outputs.tag }}.zip"
           out-file-path: "distribution/docker/release-bin"
           extract: false
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -52,7 +52,7 @@ jobs:
         run: |
           set -euo pipefail
           RELEASE_TAG="${INPUT_TAG:-$REF_NAME}"
-          if [[ ! "$RELEASE_TAG" =~ ^v[0-9]+[.][0-9]+[.][0-9]+ ]]; then
+          if [[ ! "$RELEASE_TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+ ]]; then
             echo "::error::'$RELEASE_TAG' is not a release tag. Dispatch this workflow on the tag, or pass the 'tag' input (e.g. v7.4.0)."
             exit 1
           fi
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -13,8 +13,6 @@ on:
 permissions:
   contents: write
   pull-requests: write
-  # Needed to start the 'Build Docker Image' workflow via workflow_dispatch.
-  actions: write
 
 jobs:
   build-release:
@@ -103,13 +101,34 @@ jobs:
           artifacts: "core/target/classes/com/predic8/membrane/core/config/json/membrane.schema.json,distribution/target/membrane-api-gateway-${{ steps.get_version.outputs.VERSION }}.zip,distribution/target/membrane-api-gateway-${{ steps.get_version.outputs.VERSION }}.zip.asc"
           token: ${{ github.token }}
 
-      # A release created with the default GITHUB_TOKEN does not fire the
-      # 'release: published' event: GitHub suppresses workflow triggering for
-      # events created by GITHUB_TOKEN, so the Docker build never started.
-      # workflow_dispatch is explicitly exempt from that restriction, so the
-      # build is kicked off here instead. It is dispatched on the base branch
-      # (the tag has no workflow file of its own for older releases) and gets
-      # the release tag passed as an input.
+  trigger-website:
+    needs: build-release
+    uses: ./.github/workflows/trigger-website-pipeline.yml
+    with:
+      version: ${{ needs.build-release.outputs.version }}
+    secrets: inherit
+
+  # A release created with the default GITHUB_TOKEN does not fire the
+  # 'release: published' event: GitHub suppresses workflow triggering for
+  # events created by GITHUB_TOKEN, so the Docker build never started.
+  # workflow_dispatch is explicitly exempt from that restriction, so the build
+  # is kicked off here instead.
+  #
+  # Dispatched on the base branch rather than on the tag, because release-pr.yml
+  # creates a branch AND a tag named v<version>, so '--ref v<version>' would be
+  # ambiguous; the tag is passed as an input instead.
+  #
+  # Its own job on purpose: a dispatch that fails -- against a branch whose
+  # docker-publish.yml predates the 'tag' input the API rejects it with 422 --
+  # must not fail the release itself or block the jobs beside it. It also keeps
+  # 'actions: write' off the Maven steps in build-release.
+  trigger-docker:
+    needs: build-release
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      actions: write
+    steps:
       - name: Trigger Docker image build
         shell: bash
         env:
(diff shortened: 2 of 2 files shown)


predic8 (commented):
Reviewed the workflow changes — the diagnosis is right and the fix works. I verified the mechanics independently:

* `docker/metadata-action` docs confirm `latest=auto` emits `latest` for any `type=semver` match **regardless** of the `enable=` on the raw entry, so the old condition really was dead code and `flavor: latest=false` is the correct fix. `value=` on `type=semver` is documented and supported in v4.
* The `inputs.latest` boolean is safe: the `inputs` context [preserves Boolean values as Booleans](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_dispatch) (unlike `github.event.inputs`), and the live v6.6.0 dispatch ([run 34100932880](https://github.com/membrane/api-gateway/actions/runs/34100932880)) pushed `6.6.0`/`6.6`/`6` but *not* `latest` — Docker Hub's `latest` still carries the 7.5.0 digest.
* Checking out `refs/tags/<tag>` also fixes a latent ambiguity: `release-pr.yml` creates a branch **and** a tag named `v<version>`, so `ref: ${{ github.ref }}` could previously have built from the branch.

Comments inline. One finding doesn't fit on the diff:

---

### `maven-publish.yml` has the identical broken trigger

`.github/workflows/maven-publish.yml` triggers on the same event and is suppressed for the same reason:

```yaml
on:
  release:
    types: [published]
```

Maven Central confirms the same gap as Docker Hub:

```
... 6.5.3  6.5.4  7.2.5  7.3.1  7.5.0
```

`7.4.0` and `6.6.0` are missing — so **v7.4.0 was never published to Maven Central either**, and 7.3.1/7.5.0 got there the same way their images did: by hand.

One wrinkle if the dispatch is extended to it: `maven-publish.yml` checks out `${{ github.ref }}` and runs `mvn deploy`, so dispatching it on `$BASE_REF` would deploy `7.5.1-SNAPSHOT` rather than the release. It needs the same `tag`-input treatment as `docker-publish.yml` — and dispatching it on the tag name isn't safe either, for the `v7.4.0`-is-both-a-branch-and-a-tag reason above. Merging this PR alone leaves the release process half-fixed while looking fixed.

---

CodeRabbit's two points (complete SemVer match, `metadata-action@v4` requesting the removed Node 16 runtime) are both legitimate. On the second: `v4` isn't alone — `docker/setup-qemu-action@v2`, `docker/build-push-action@v4` and the 2022 `setup-buildx-action` SHA pin are all in the same file, so a separate bump is due before 23 Sep 2026.

.github/workflows/docker-publish.yml
```diff
@@ -69,21 +94,29 @@ jobs:
...
+          flavor: |
+            latest=false
           tags: |
-            type=semver,pattern={{version}}
-            type=semver,pattern={{major}}.{{minor}}
-            type=semver,pattern={{major}}
-            type=raw,value=latest,enable=${{ ! github.event.release.prerelease }}
+            type=semver,pattern={{version}},value=${{ steps.release_version.outputs.release_tag }}
+            type=semver,pattern={{major}}.{{minor}},value=${{ steps.release_version.outputs.release_tag }}
+            type=semver,pattern={{major}},value=${{ steps.release_version.outputs.release_tag }}
```
predic8, line 113: **The floating `7` / `7.4` tags are still unguarded on a backfill.**

`latest` is now carefully gated, but the other moving tags aren't — and this PR is precisely what makes backfilling an old release easy (the description proposes doing it for v7.4.0).

Docker Hub right now:

```
7        -> sha256:7b7fbb3b...  (= 7.5.0)
7.5      -> sha256:7b7fbb3b...
7.5.0    -> sha256:7b7fbb3b...
```

Dispatching `tag=v7.4.0` today emits `7.4.0`, `7.4` **and `7`**, silently repointing `7` from 7.5.0 back to 7.4.0. Same class of defect as the `latest` one this PR fixes, one step less visible.

Either gate `{{major}}`/`{{major}}.{{minor}}` behind an input alongside `latest`, or at minimum note it in the `tag` input description so whoever backfills v7.4.0 knows what else moves.

.github/workflows/docker-publish.yml
```diff
@@ -9,7 +9,18 @@ on:
...
+      tag:
+        description: "Release tag to build, e.g. v7.4.0. Required unless the workflow is run on the tag itself."
+        type: string
+        required: false
+        default: ''
+      latest:
+        description: "Also push the 'latest' Docker tag"
+        type: boolean
+        required: false
+        default: false
```
predic8, line 22: **Consider dropping this input in favour of the existing `update-docker-latest.yml`.**

`distribution/docker/Dockerfile` runs `apt-get update && apt-get upgrade -y`, so a rebuild is not reproducible: re-dispatching an already-published tag to move `latest` also replaces the published `7.5.0` bits with different ones under the same tag.

`update-docker-latest.yml` ("Promote Docker Tag to Latest") already does this with `docker buildx imagetools create` — no rebuild, identical digest, no drift. That's the safer path for what the v7.5.0 dispatch did manually, and dropping the input reduces the `enable=` expression below to just the release-event half.

.github/workflows/docker-publish.yml
```diff
@@ -29,16 +40,30 @@ jobs:
...
+          REF_NAME: ${{ github.ref_name }}
+        run: |
+          set -euo pipefail
+          RELEASE_TAG="${INPUT_TAG:-$REF_NAME}"
+          if [[ ! "$RELEASE_TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+ ]]; then
+            echo "::error::'$RELEASE_TAG' is not a release tag. Dispatch this workflow on the tag, or pass the 'tag' input (e.g. v7.4.0)."
+            exit 1
+          fi
+          echo "release_tag=$RELEASE_TAG" >> "$GITHUB_OUTPUT"
+          echo "tag=${RELEASE_TAG#v}" >> "$GITHUB_OUTPUT"
```
predic8, line 60: Nit: three overlapping names for two values — the workflow input `tag` is `v7.4.0`, `outputs.release_tag` is `v7.4.0`, and `outputs.tag` is `7.4.0`. `tag` meaning the version *without* the `v` while `inputs.tag` means the version *with* it is the confusing part; it only survives from the removed substring action's `output_name`. Renaming this one to `version` costs one more line at the `release-downloader` step.

.github/workflows/release-build.yml
```diff
@@ -101,6 +103,28 @@ jobs:
...
+          REPO: ${{ github.repository }}
+          BASE_REF: ${{ github.base_ref }}
+          RELEASE_TAG: ${{ github.head_ref }}
+        run: |
+          set -euo pipefail
+          gh workflow run docker-publish.yml \
+            --repo "$REPO" \
+            --ref "$BASE_REF" \
+            --field tag="$RELEASE_TAG" \
+            --field latest=false
```
predic8, line 145: **A failed dispatch now fails the release job and skips the website pipeline.**

This is the last step of `build-release`, and `trigger-website` declares `needs: build-release`. If `gh workflow run` fails — a release cut from `6.X` before #3213 lands, or from `5**`/`4**`, which are still in the trigger list and whose `docker-publish.yml` has no `tag` input (the dispatch API 422s on unexpected inputs) — the result is: release created, artifacts uploaded, job red, **website pipeline skipped**. Before this PR the Docker build merely didn't happen.

Its own job decouples that:

```yaml
  trigger-docker:
    needs: build-release
    runs-on: ubuntu-latest
    permissions:
      actions: write
    steps: ...
```

That also scopes the permission. As written, `actions: write` applies to the whole workflow, including the `mvn install` / `mvn verify` steps that execute third-party plugin code — minor next to the existing `contents: write`, but free to avoid.

.github/workflows/release-build.yml
```diff
@@ -101,6 +103,28 @@ jobs:
           artifacts: "core/target/classes/com/predic8/membrane/core/config/json/membrane.schema.json,distribution/target/membrane-api-gateway-${{ steps.get_version.outputs.VERSION }}.zip,distribution/target/membrane-api-gateway-${{ steps.get_version.outputs.VERSION }}.zip.asc"
           token: ${{ github.token }}
 
+      # A release created with the default GITHUB_TOKEN does not fire the
+      # 'release: published' event: GitHub suppresses workflow triggering for
+      # events created by GITHUB_TOKEN, so the Docker build never started.
+      # workflow_dispatch is explicitly exempt from that restriction, so the
+      # build is kicked off here instead. It is dispatched on the base branch
+      # (the tag has no workflow file of its own for older releases) and gets
+      # the release tag passed as an input.
```
predic8: "the tag has no workflow file of its own for older releases" isn't quite right — the tag does carry `.github/workflows/docker-publish.yml`, just an older version of it without the `tag` input.

The PR description's reason is the better one and the one worth having here, since it's the constraint a future reader would otherwise break: `release-pr.yml` creates both a branch and a tag named `v<version>`, so `--ref v<version>` is ambiguous.

rrayst:
Thanks — three responses, one of them a correction that matters.

### Correction: the v6.6.0 run *did* push `latest`

> the live v6.6.0 dispatch ([run 34100932880](https://github.com/membrane/api-gateway/actions/runs/34100932880)) pushed `6.6.0`/`6.6`/`6` but *not* `latest`

It pushed `latest` as well. That run executed on commit 3a2d226, which is **before** the `flavor: latest=false` commit, and its metadata step logged:

```
##[group]Docker tags
docker.io/***/membrane:6.6.0
docker.io/***/membrane:6.6
docker.io/***/membrane:6
docker.io/***/membrane:latest
##[endgroup]
```

`predic8/membrane:latest` then carried 6.6.0's digest (`sha256:3ff4ea80…`) and was repointed to 7.5.0 (`sha256:7b7fbb3b…`) by hand afterwards. It is correct again now.

That run is therefore evidence of the bug, not of the fix — **`flavor: latest=false` has not been exercised by any run yet.** Worth confirming on the next release from the metadata step's log: the `Processing flavor input` group should read `latest=false`, and the `Docker tags` group should have no `latest` line.

### CodeRabbit's SemVer anchor — applied, with a change

Valid, but the literal suggestion (`$` straight after the last `[0-9]+`) would reject `v7.0.0-RC1`, which is a real tag in this repo. Anchored the whole string while allowing the optional prerelease and build-metadata parts instead:

```bash
^v[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?([+][0-9A-Za-z.-]+)?$
```

Checked against `v6.6.0`, `v7.4.0`, `v7.0.0-RC1`, `v6.3.15`, `v7.4.0+build.5` (all accepted) and `v7.4.0junk`, `v7.4.0.1`, `v7.4`, `master`, `6.X` (all rejected). Pushed here and to #3213.

### CodeRabbit's `metadata-action@v4` / Node 16 point — correct, but not specific to that action

The runtime claim holds (`action.yml`: v4=`node16`, v5=`node20`, v6=`node24`), but it isn't alone. Every non-checkout action in this file is node16:

| action | runtime |
|---|---|
| `docker/metadata-action@v4` | node16 |
| `docker/setup-qemu-action@v2` | node16 |
| `docker/setup-buildx-action` (2022 SHA pin) | node16 |
| `docker/build-push-action@v4` | node16 |
| `docker/login-action@v2` | node16 |
| `robinraju/release-downloader@v1.8` | node16 |
| `peter-evans/dockerhub-description@v3` | node16 |
| `actions/checkout@v7` | node24 |

Bumping one of eight buys nothing, and folding an action upgrade into a release-critical logic change leaves two suspects if the next release breaks. Handling all of them in a separate PR.

### On `maven-publish.yml`

Confirmed missing on Central — `repo1.maven.org/maven2/org/membrane-soa/service-proxy-core` lists `6.5.3, 6.5.4, 7.2.5, 7.3.1, 7.5.0`, so 7.4.0 and 6.6.0 are absent, and the `${{ github.ref }}` + `mvn deploy` wrinkle is real.

One qualification on "leaves the release process half-fixed": the release-triggered Maven runs were **already failing before the token change** — v7.2.3 (Jun 10), v7.2.4 (Jun 25), v7.2.5 and v6.5.3 (Jun 26) all ended with

```
[ERROR] Unable to upload bundle for deployment: Deployment
[ERROR] ... central-publishing-maven-plugin:0.8.0:publish ... failed
```

Every successful Central publish since has been a manual dispatch, i.e. a retry of that upload. So Central has been a manual step for months for a reason unrelated to the trigger. Wiring the dispatch in turns a suppressed trigger into an automatically failing one unless the upload flakiness is addressed too — separate PR, noting that caveat.

.github/workflows/docker-publish.yml
rrayst: "reproducible" was never defined as goal.

.github/workflows/docker-publish.yml
rrayst: the version number is clear.

.github/workflows/docker-publish.yml
rrayst: please open an issue, if required.

.github/workflows/docker-publish.yml
```diff
@@ -69,21 +94,29 @@ jobs:
...
       # https://github.com/docker/metadata-action
+      # 'latest' follows the release for a real release, and the 'latest' input
+      # for a manual run. Without the explicit event check a manual run would
+      # always push 'latest', because github.event.release is empty there.
+      # 'flavor: latest=false' is required as well: metadata-action defaults to
+      # 'latest=auto', which adds 'latest' for any semver version regardless of
+      # the enable= condition on the raw tag below.
       - name: Extract Docker metadata
         id: meta
         uses: docker/metadata-action@v4
```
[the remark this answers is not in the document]
rrayst: -> #3225

.github/workflows/release-build.yml
rrayst: Fixed — the dispatch is its own `trigger-docker` job now, `needs: build-release`, with `contents: read` + `actions: write` scoped to it, and `actions: write` removed from the workflow-level block.

Worth recording that on `6.X` (#3213) it was worse than described: there the GitLab website trigger is still an inline step in `build-release`, and the dispatch sat directly *before* it — so a failed dispatch skipped the website pipeline outright rather than via `needs:`. Same fix applied on both branches.

The 5**/4** case is real: those branches are in the trigger list and their `docker-publish.yml` has no `tag` input, so the dispatch API 422s.

.github/workflows/release-build.yml
rrayst: Correct — the tag does carry `docker-publish.yml`, just an older copy without the `tag` input. Comment rewritten to give the constraint that actually binds:

```
  # Dispatched on the base branch rather than on the tag, because release-pr.yml
  # creates a branch AND a tag named v<version>, so '--ref v<version>' would be
  # ambiguous; the tag is passed as an input instead.
```

Applied on both branches.

predic8 (approved):

PR #3227, opened 2026-09-07
fix(cli): harden option parsing, validation and help across all subcommands

predic8:
Closes #3224
Closes #3219

## Summary by CodeRabbit

* **New Features**
  * Added consistent `-h`/`--help` support across CLI commands, including when required options are omitted.
  * Added validation for numeric command-line values, with supported ranges and defaults.
  * RSA key generation now defaults to 2048 bits and accepts sizes from 2048 to 16384 bits.

* **Bug Fixes**
  * Improved handling of whitespace in command-line values.
  * Argon2id now accepts only supported versions 16 and 19.

* **Documentation**
  * Help text now documents valid value ranges and defaults.

Issue #3222, opened 2026-09-07
CLI: no `-h`/`--help` for `generate-jwk`, `private-jwk-to-public` and `argon2id` — `argon2id -h` prompts for a password


predic8:
The root command and `oas` declare a `-h` / `--help` option, but `generate-jwk`, `private-jwk-to-public` and `argon2id` do not. Asking them for help therefore produces something else — in the case of `argon2id`, an interactive password prompt that blocks the terminal.

### Reproduce

```
$ membrane.sh argon2id -h
Enter password to hash:
<hangs; with stdin closed: "Exception: NoSuchElementException: No line found">

$ membrane.sh private-jwk-to-public -h
ERROR Both input (-i) and output (-o) files must be specified.

$ membrane.sh generate-jwk -h
usage: membrane.sh generate-jwk [options]     # only as a side effect of "Missing required option: o", exit code 1
```

### Cause

`RouterCLI.start` checks `isOptionSet("h")` (`RouterCLI.java:79`), but the three subcommands never declare the option (`MembraneCommandLine.java:61-79`), so it is always `false` for them and execution falls through to the command body. In `argon2id` that body reads the password from `System.in` (`RouterCLI.java:132-136`).

### Suggested fix

Add `builder("h").longOpt("help")` to every subcommand — or, better, register it once in `CliCommand` so no subcommand can forget it. `generate-jwk -h` should also exit with 0 rather than 1.

Found in a review of the `com.predic8.membrane.core.cli` package (master 24fc06ce6, 7.5.1-SNAPSHOT).

Issue #3221, opened 2026-09-07
CLI: `--overwrite` is not recognized by `generate-jwk` (only `-overwrite`)


predic8:
The `overwrite` flag of `generate-jwk` is declared without a long option, so only `-overwrite` works. `--overwrite` — the spelling every other option in the CLI supports — is silently ignored.

### Reproduce

```
$ membrane.sh generate-jwk -o key.json
$ membrane.sh generate-jwk -o key.json --overwrite
ERROR Output file (key.json) already exists.

$ membrane.sh generate-jwk --overwrite -o key.json
usage: membrane.sh generate-jwk [options]      # "Missing required option: o"

$ membrane.sh generate-jwk -o key.json -overwrite
# works
```

### Cause

```java
// MembraneCommandLine.java:64
addOption(builder("overwrite").desc("Overwrite the output file, if it exists.").build());
```

No `.longOpt("overwrite")`, so `--overwrite` is an unknown token. Because the parser runs with `stopAtNonOption=true` it is not reported as an error but swallows the remaining arguments (#3218), which is why the second invocation loses `-o` as well.

### Suggested fix

Declare it like the other flags: `builder("overwrite").longOpt("overwrite")...`, or give it a short name plus a long option.

Found in a review of the `com.predic8.membrane.core.cli` package (master 24fc06ce6, 7.5.1-SNAPSHOT).

Issue #3220, opened 2026-09-07
CLI: option values are trimmed — `argon2id` hashes a different password than given


predic8:
`CliCommand.getOptionValue` trims every option value, so an `argon2id` password with leading or trailing whitespace is hashed as if the whitespace were not there. The resulting PCH can never be verified against the password the user actually chose.

### Reproduce

Two different passwords, one identical hash:

```
$ membrane.sh argon2id -pass "  abc  " -s 00112233445566778899aabbccddeeff
$argon2id$v=19$m=65536,t=3,p=1$ABEiM0RVZneImaq7zN3u/w$ejCVAj/rnFsJiS16hAmvyLN8rT/l3Cv3nMq9olHKTKg

$ membrane.sh argon2id -pass "abc" -s 00112233445566778899aabbccddeeff
$argon2id$v=19$m=65536,t=3,p=1$ABEiM0RVZneImaq7zN3u/w$ejCVAj/rnFsJiS16hAmvyLN8rT/l3Cv3nMq9olHKTKg
```

A user who hashes `"secret "` and puts the result into an `apis.yaml` will be rejected by `SecurityUtils.verifyArgon2id` when logging in with `"secret "`.

### Cause

```java
// CliCommand.java:150
public String getOptionValue(String opt) {
    return commandLine != null ? trim(commandLine.getOptionValue(opt)) : null;
}
```

`org.apache.commons.lang3.StringUtils.trim` is applied to *all* option values, credentials included.

### Suggested fix

Don't trim in the generic accessor. If trimming is wanted for file locations, do it at those call sites; option values that carry secrets (`-pass`, `-s`) must be passed through byte-for-byte.

Found in a review of the `com.predic8.membrane.core.cli` package (master 24fc06ce6, 7.5.1-SNAPSHOT).

predic8:
Fixed in #3227.

`CliCommand.getOptionValue` now returns the option value byte-for-byte. The whitespace tolerance that commit 21e1a8ef6 introduced for config paths moved into a separate `getTrimmedOptionValue`, used for the location-like options (`-c`, `-t`, `-l`, `-i`, `-o`, and the argon2id salt, which is hex-decoded anyway) and inside `getIntOptionValue`.

[...]

PR #3228, opened 2026-09-07
fix(cli): harden option parsing, validation and help across all subcommands

### LINKED ISSUE #3216: CLI: `oas` subcommand always fails with NPE — port is opened before the transport exists
`membrane.sh oas -l <location>` always fails with a `NullPointerException`, so the whole subcommand is unusable — including both `oas` examples printed by `membrane.sh --help`.

### Reproduce

```
$ membrane.sh oas -l conf/fruitshop-api.yml
java.lang.NullPointerException: Cannot invoke "com.predic8.membrane.core.transport.Transport.openPort(com.predic8.membrane.core.proxies.SSLableProxy)" because the return value of "com.predic8.membrane.core.router.Router.getTransport()" is null
	at com.predic8.membrane.core.proxies.RuleManager.addProxyAndOpenPortIfNew(RuleManager.java:81)
	at com.predic8.membrane.core.proxies.RuleManager.addProxyAndOpenPortIfNew(RuleManager.java:74)
	at com.predic8.membrane.core.cli.RouterCLI.initRouterByOpenApiSpec(RouterCLI.java:234)
	at com.predic8.membrane.core.cli.RouterCLI.getRouter(RouterCLI.java:187)
	at com.predic8.membrane.core.cli.RouterCLI.start(RouterCLI.java:102)
```

Happens with a local file and with a URL.

### Cause

`RouterCLI.initRouterByOpenApiSpec` opens the port before the router is initialized:

```java
var router = new DefaultRouter();
router.getRuleManager().addProxyAndOpenPortIfNew(getApiProxy(commandLine)); // needs the transport
router.init();                                                             // creates the transport
```

`RuleManager.addProxyAndOpenPortIfNew` calls `router.getTransport().openPort(proxy)`, but the transport is only created in `DefaultMainComponents.init()` (`if (transport == null) transport = new HttpTransport();`), which runs one line later.

The order used to be harmless: the method was written against `HttpRouter`, which built its transport in the constructor. It broke when the CLI switched to `DefaultRouter` (#2514 / #2523).

### Second problem in the same method

`init()` does not set `running`, so even with the NPE fixed `DefaultRouter.waitFor()` (`RouterCLI.java:103`) returns immediately, and JMX export, the hot deployer and the rule reinitializer are never started — they all live in `start()`.

### Suggested fix

Mirror `initRouterByYAML`: register the proxy, then let `start()` do init + `openPorts()`:

```java
var router = new DefaultRouter();
router.getRuleManager().addProxy(getApiProxy(commandLine), MANUAL);
router.start();
logInfosAboutStartedProxies(router.getRuleManager());
logStartupMessage();
return router;
```

There is currently no test covering the `oas` path.

Related: #2338 (same subcommand, different failure mode, closed 2025-11-21).

Found in a review of the `com.predic8.membrane.core.cli` package (master 24fc06ce6, 7.5.1-SNAPSHOT).

### LINKED ISSUE #3219: CLI: `private-jwk-to-public -i k.json -o k.json` destroys the private key
`private-jwk-to-public` writes its output file unconditionally. Pointing `-o` at the input file silently replaces the private key with its public half — the private key material is gone and cannot be recovered.

### Reproduce

```
$ membrane.sh generate-jwk -o key.json
$ python3 -c "import json;print(sorted(json.load(open('key.json'))))"
['alg', 'd', 'dp', 'dq', 'e', 'kid', 'kty', 'n', 'p', 'q', 'qi', 'use']

$ membrane.sh private-jwk-to-public -i key.json -o key.json    # no warning, exit 0
$ python3 -c "import json;print(sorted(json.load(open('key.json'))))"
['alg', 'e', 'kid', 'kty', 'n', 'use']
```

### Cause

`JwkGenerator.privateJWKtoPublic` (`util/JwkGenerator.java:81`) reads the input, then calls `Files.writeString(get(output), ...)`, which truncates:

```java
Map map = new ObjectMapper().readValue(new File(input), Map.class);
RsaJsonWebKey rsa = new RsaJsonWebKey(map);
writeString(get(output), rsa.toJson(PUBLIC_ONLY));
```

There is no existing-file check (unlike `generate-jwk`, which refuses to overwrite without `-overwrite`) and no input/output comparison.

### Suggested fix

Refuse to write when the output file already exists unless an explicit `-overwrite` is given — consistent with `generate-jwk` — and reject `-i` and `-o` resolving to the same file outright.

`-i` and `-o` are also not declared `required()` (`MembraneCommandLine.java:68-69`); the null check in `RouterCLI.privateJwkToPublic` compensates for that at runtime.

Found in a review of the `com.predic8.membrane.core.cli` package (master 24fc06ce6, 7.5.1-SNAPSHOT).

### LINKED ISSUE #3224: CLI: invalid numeric options print a raw stack trace (`generate-jwk -b`, `oas -p`)
Numeric command line options are parsed with `Integer.parseInt` without handling `NumberFormatException`, so a typo produces a Java stack trace instead of an error message.

### Reproduce

```
$ membrane.sh generate-jwk -o key.json -b abc
Exception in thread "main" java.lang.NumberFormatException: For input string: "abc"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)
	at java.base/java.lang.Integer.parseInt(Integer.java:565)
	at java.base/java.lang.Integer.parseInt(Integer.java:662)
	at com.predic8.membrane.core.cli.util.JwkGenerator.generateJWK(JwkGenerator.java:46)
	...

$ membrane.sh oas -l api.yml -p abc
ERROR …NumberFormatException: For input string: "abc"
	...
	at com.predic8.membrane.core.cli.RouterCLI.getRouter(RouterCLI.java:187)
	at com.predic8.membrane.core.cli.RouterCLI.start(RouterCLI.java:102)
	at com.predic8.membrane.core.cli.RouterCLI.main(RouterCLI.java:68)
```

Related, same shape: `membrane.sh oas -l ""` logs `ERROR null` plus the stack trace of a message-less `RuntimeException` thrown by the guard at `RouterCLI.java:276` (which is commented "unreachable" — `-l` is `required()`, but that only guarantees presence, not a non-empty value).

### Suggested fix

Validate the numeric options where they are read and print a one-line message, e.g. `Invalid value for -b: 'abc' is not a number.` `argon2id` already wraps its `parseInt` calls in a `try`/`catch` (`RouterCLI.java:119-148`) and prints a message — the other two paths should behave the same. A plausible range check (`-b` >= 2048, `-p` a valid port) would be worth adding at the same time.

Found in a review of the `com.predic8.membrane.core.cli` package (master 24fc06ce6, 7.5.1-SNAPSHOT).

predic8:
Closes #3224
Closes #3219

Supersedes #3227, which GitHub closed when the head branch was renamed to `fix/cli-argument-hardening`. Same commits, same base.

Hardens CLI argument handling across every subcommand:

- **Option values are validated.** Numeric options are type- and range-checked (`getIntOptionValue`), reporting the offending value instead of an unhandled `NumberFormatException`. `argon2id -v` accepts only the two versions Argon2 defines (16, 19); `oas -p` is limited to 1-65535; `generate-jwk -b` to 2048-16384.
- **Nothing is silently discarded.** Unknown options and unexpected positional arguments now abort with a message and the failing command's help. Previously the parser stopped at the first unknown token, so `--typo -c apis.yaml` dropped the config and started with a different configuration.
- **`-h` works everywhere**, including on commands that declare no options of their own and when a required option is missing.
- **Key material is protected.** `private-jwk-to-public` refuses to write onto its own input (which used to discard the private key components), requires `-i`/`-o`, and needs `-overwrite` to replace an existing output.
- **Values reach their consumer byte-for-byte.** Option values are no longer trimmed by default, so an `argon2id` password with leading or trailing whitespace hashes correctly; file locations opt in via `getTrimmedOptionValue`.
- **`start -t` performs a dry run**, as its "same function as command omitted" documentation promises.
- **`oas -l` starts the router through the documented path** (`addProxy` + `start()`), so OpenAPI specs are validated at startup instead of failing on a port opened before the transport existed.

Also fixes #3216, #3217, #3218, #3220, #3221 and #3222 (not listed as `Closes` - promote them if they should auto-close on merge).

## Summary by CodeRabbit

- **New Features**
  - Added consistent help support across command-line commands, including commands with required options.
  - Added validated integer option handling and clearer command-specific error reporting.
  - Added overwrite controls for JWK generation and conversion.
  - Added validation for ports, Argon2 versions, OpenAPI locations, and file paths.

- **Bug Fixes**
  - Unexpected arguments and unknown commands or options are now rejected.
  - Dry-run mode works correctly with the start command.
  - JWK operations prevent unintended file replacement and protect private-key files.
  - CLI values are trimmed consistently where appropriate.


core/src/main/java/com/predic8/membrane/core/cli/util/JwkGenerator.java
```diff
@@ -24,36 +25,48 @@
...
-    public static void generateJWK(MembraneCommandLine commandLine) {
-        int bits = 2048;
-        String bitsArg = commandLine.getCommand().getOptionValue("b");
-        if (bitsArg != null) {
-            bits = parseInt(bitsArg);
-        }
+    /**
+     * Other users must not be able to read a private key.
+     */
+    private static final Set<PosixFilePermission> OWNER_ONLY = fromString("rw-------");
```
rrayst, line 56: This breaks backward compatibility. Should be highlighted in the CHANGELOG.

core/src/main/java/com/predic8/membrane/core/cli/RouterCLI.java
```diff
@@ -229,17 +254,17 @@ private static Router initRouterByConfig(MembraneCommandLine commandLine) throws
         throw new RuntimeException("Unsupported file extension.");
     }
 
-    private static Router initRouterByOpenApiSpec(MembraneCommandLine commandLine) throws Exception {
+    static Router initRouterByOpenApiSpec(MembraneCommandLine commandLine) throws Exception {
         var router = new DefaultRouter();
-        router.getRuleManager().addProxyAndOpenPortIfNew(getApiProxy(commandLine));
-        router.init();
+        router.getRuleManager().addProxy(getApiProxy(commandLine), MANUAL);
```
rrayst, line 259: `MANUAL` is the wrong value. (see comment on the constant)

rrayst: use `SPRING` (or rename that)

rrayst (changes requested):

rrayst (approved):

Issue #3238, opened 2026-09-08
Concurrent writes to an externally stored session lose updates, breaking OAuth2 CSRF state


rrayst:
Concurrent requests on one session lose each other's writes when the session lives in an external key-value store (`memcachedSessionManager`, `redisSessionManager`). With OAuth2 client flows this drops CSRF state tokens and the callback is rejected with `400 CSRF token mismatch.`

## Mechanism

For externally stored sessions each request gets its **own deserialized copy** of the session and writes the whole thing back with a blind SET. There is no CAS, so the last writer wins and every concurrent writer's changes are lost.

`MemcachedSessionManager`:

```java
    protected Map<String, Object> cookieValueToAttributes(String cookie) {
        return getCachedSession(cookie).map(this::parse)          // fresh copy per request
                .orElse(new Session(usernameKeyName, new HashMap<>())).get();
    }

    protected void addSessions(Session[] sessions) {
        Arrays.stream(sessions).forEach(s -> {
            client.set(s.get(ID_NAME), ..., stringify(s));         // blind SET, no CAS
```

`RedisSessionManager` has the same shape — `jsonStringtoSession(jedis.getEx(...)).get()` on read, `jedis.setex(...)` on write.

`InMemorySessionManager` is unaffected, and the contrast is the point: its `cookieValueToAttributes` returns the **live map** of the shared `Session` held in its Guava cache, so concurrent requests mutate one object and nothing is lost.

## Why OAuth2 hits it

`StateManager.saveToSession` is a read-modify-write that appends to a multi-valued `state` key, explicitly so several in-flight authorization flows can coexist:

```java
    public void saveToSession(Session session) {
        String s = securityToken;
        if (session.get().containsKey(SESSION_PARAMETER_STATE))
            s = session.get(SESSION_PARAMETER_STATE) + SESSION_VALUE_SEPARATOR + s;
        session.put(SESSION_PARAMETER_STATE, s);
    }
```

That merge silently assumes the read-modify-write is atomic. It is with `inMemorySessionManager2`; it is not with an external store. Two concurrent flows both read `state=t0`, one writes `t0,tA` and the other `t0,tB`, and whichever lands second wins. The request whose token was dropped then fails `StateManager.verifyCsrfToken` — session is not new and `hasState` is true, so it takes the mismatch branch and returns `badRequest().body("CSRF token mismatch.")`.

[...]

PR #3239, opened 2026-09-08
Share one HttpClient per BrowserMock instead of building one per request

rrayst:
Test-scope leak fix. Stacked on #3237, which introduces the `ConnectRetry` helper this touches; merge that one first and this retargets to `master` automatically.

> Verification was still running when this was opened — results will be posted as a comment.

## The leak

`BrowserMock` routed every request through

```java
    final Function<Exchange, Exchange> cookieHandlingHttpClient = exc -> cookieManager(httpClient(), exc);
```

and `httpClient()` constructed a **new `HttpClient` on every call**, i.e. one per HTTP request, never closed:

```java
    private Function<Exchange, Exchange> httpClient() {
        return new Function<>() {
            final HttpClient httpClient = new HttpClient(getHttpClientConfiguration());
            ...
```

Each of those carries its own `ConnectionFactory` → `ConnectionManager`, and since no `TimerManager` is passed, the `ConnectionManager` creates its own `Timer` thread:

```java
        var autoCloseInterval = keepAliveTimeout * 2;
        if (timerManager == null)
            selfCreatedTimerManager = timerManager = new TimerManager();
        timerManager.schedulePeriodicTask(..., autoCloseInterval, "Connection Closer");
```

It also holds pooled keep-alive connections for `keepAliveTimeout` (4 s by default). The B2C tests drive 500 workers through a redirect, form-post and token-refresh chain, so one test run leaked thousands of clients, sockets and timer threads.

## The fix

Hold one `HttpClient` per `BrowserMock` and reuse it — which is how the gateway itself uses the class: a single instance serves every request thread. The anonymous `Function` collapses into a private `call(Exchange)` method with the same body and the same `RuntimeException("while calling " + ...)` wrapping.

`BrowserMock` becomes `AutoCloseable`, and its four users close it in their `@AfterEach`, so the one remaining client and its timer go away with the test: `OAuth2ResourceTest`, `OAuth2ResourceErrorForwardingTest`, `OAuth2ResourceRpIniLogoutTest`, `OAuth2ResourceB2CTestSetup`.

## Not the cause of the Windows failures

This was found while diagnosing #3237 and was explicitly cleared of causing those failures: every lost request there was a `ConnectException` from the accept queue Windows caps at 200, with no `BindException` and no thread-creation failure anywhere in the evidence. It is a leak worth fixing on its own terms, which is why it is not bundled into #3237.

## Review focus

The interesting risk is the opposite of the leak: a single `HttpClient` is now shared across 500 concurrent workers, and `close()` calls `shutdownWhenDone()` on the connection manager. Verification therefore targets

- `InMemB2CResourceIntegrationTest` and `JwtB2CResourceIntegrationTest` under full 500-way concurrency, watching for any new pool or keep-alive exception type
- the three non-B2C `BrowserMock` users, whose teardown changed
- `InMemB2CResourceTest` as a second regression check

Note that `SyncB2CResourceIntegrationTest` fails in roughly half of runs for an unrelated reason tracked in #3238 (lost session updates breaking OAuth2 CSRF state); a CSRF failure there is not attributable to this change.

## Summary by CodeRabbit

- **Tests**
  - Improved OAuth2 test cleanup by consistently releasing browser resources after each test.
  - Reused a shared HTTP client during browser-mock interactions, reducing repeated client creation.
  - Added automatic resource closing support to the test browser utility.

### DIFF
--- a/core/src/test/java/com/predic8/membrane/core/interceptor/oauth2/client/BrowserMock.java
+++ b/core/src/test/java/com/predic8/membrane/core/interceptor/oauth2/client/BrowserMock.java
@@ -33,13 +33,14 @@ import static java.nio.charset.StandardCharsets.*;
 import static java.util.stream.Collectors.*;
 import static org.apache.commons.text.StringEscapeUtils.*;
 
-public class BrowserMock implements Function<Exchange, Exchange> {
+public class BrowserMock implements Function<Exchange, Exchange>, AutoCloseable {
 
     public static final Pattern INPUT_PATTERN = Pattern.compile("<input type=\"hidden\" name=\"([-a-zA-Z0-9&;_]*)\" value=\"([-a-zA-Z ._~=0-9&/;]*)\"/>");
     public static final Pattern FORM_PATTERN = Pattern.compile("<form method=\"post\" action=\"([a-z:/0-9]*)\"");
     final Logger LOG = LoggerFactory.getLogger(BrowserMock.class);
     final Map<String, Map<String, String>> cookie = new HashMap<>();
-    final Function<Exchange, Exchange> cookieHandlingHttpClient = exc -> cookieManager(httpClient(), exc);
+    final HttpClient httpClient = new HttpClient(getHttpClientConfiguration());
+    final Function<Exchange, Exchange> cookieHandlingHttpClient = exc -> cookieManager(this::call, exc);
     final Function<Exchange, Exchange> cookieHandlingRedirectingHttpClient = outerExc -> handleFormPost(innerExc -> handleRedirect(cookieHandlingHttpClient, innerExc, new ArrayList<>()), outerExc);
 
     /**
@@ -350,20 +351,18 @@ public class BrowserMock implements Function<Exchange, Exchange> {
         return location;
     }
 
-    private Function<Exchange, Exchange> httpClient() {
-        return new Function<>() {
-            final HttpClient httpClient = new HttpClient(getHttpClientConfiguration());
+    private Exchange call(Exchange exchange) {
+        try {
+            ConnectRetry.call(httpClient, exchange);
+            return exchange;
+        } catch (Exception e) {
+            throw new RuntimeException("while calling " + exchange.getOriginalRelativeURI(), e);
+        }
+    }
 
-            @Override
-            public Exchange apply(Exchange exchange) {
-                try {
-                    ConnectRetry.call(httpClient, exchange);
-                    return exchange;
-                } catch (Exception e) {
-                    throw new RuntimeException("while calling " + exchange.getOriginalRelativeURI(), e);
-                }
-            }
-        };
+    @Override
+    public void close() {
+        httpClient.close();
     }
 
     private static @NotNull HttpClientConfiguration getHttpClientConfiguration() {
--- a/core/src/test/java/com/predic8/membrane/core/interceptor/oauth2/client/OAuth2ResourceErrorForwardingTest.java
+++ b/core/src/test/java/com/predic8/membrane/core/interceptor/oauth2/client/OAuth2ResourceErrorForwardingTest.java
@@ -82,6 +82,7 @@ public class OAuth2ResourceErrorForwardingTest {
 
     @AfterEach
     public void done() {
+        browser.close();
         if (mockAuthServer != null)
             mockAuthServer.stop();
         if (oauth2Resource != null)
--- a/core/src/test/java/com/predic8/membrane/core/interceptor/oauth2/client/OAuth2ResourceRpIniLogoutTest.java
+++ b/core/src/test/java/com/predic8/membrane/core/interceptor/oauth2/client/OAuth2ResourceRpIniLogoutTest.java
@@ -94,6 +94,7 @@ public class OAuth2ResourceRpIniLogoutTest {
 
     @AfterEach
     public void done() {
+        browser.close();
         if (mockAuthServer != null)
             mockAuthServer.stop();
         if (oauth2Resource != null)
--- a/core/src/test/java/com/predic8/membrane/core/interceptor/oauth2/client/OAuth2ResourceTest.java
+++ b/core/src/test/java/com/predic8/membrane/core/interceptor/oauth2/client/OAuth2ResourceTest.java
@@ -83,6 +83,7 @@ public abstract class OAuth2ResourceTest {
 
     @AfterEach
     public void done() {
+        browser.close();
         if (mockAuthServer != null)
             mockAuthServer.stop();
         if (oauth2Resource != null)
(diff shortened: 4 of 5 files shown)


rrayst:
Verification results, run on Windows 11 against exactly `f2ffeb888` (`git diff HEAD` empty at the time of the run).

`core` test sources compile: 612 sources, no errors and no new warnings.

## Shared client under 500 concurrent workers

| test | runs | result |
|---|---|---|
| `InMemB2CResourceIntegrationTest` | 4 (INFO) | **4/4 pass** (3154–3384 ms) |
| `InMemB2CResourceIntegrationTest` | 2 (DEBUG) | **2/2 pass** (4012, 4070 ms) |
| `JwtB2CResourceIntegrationTest` | 3 (INFO) | **3/3 pass** (2816–3375 ms) |

`java.net.ConnectException: Connection refused: getsockopt` remains the only throwable type anywhere — zero `NoResponseException`, zero `Connection reset`, zero `EOFException`, no pool-shutdown or `IllegalState` messages, no WARN or ERROR lines.

The pool is demonstrably being hammered and holding up. From a DEBUG run: 2416 × `connection requested for localhost`, 439 × `Opened connection on localPort`, 1303 × `Closing HTTP connection`, and `Number of connections in pool:` peaking at **208–251** — hundreds of live pooled connections on a single `HttpClient` across 500 threads, with 1904 refused connects all recovered and every request completing.

## The `close()`-in-teardown risk

Structurally excluded, not just unobserved: `browser` is a non-static `final` instance field in all four users, and there is no `@TestInstance(PER_CLASS)` anywhere under `core/src/test/.../oauth2/`, so JUnit's default per-method lifecycle builds a fresh `BrowserMock` — and a fresh `HttpClient` — for every test method. `close()` therefore always closes a client that is never touched again, and it is the graceful `shutdownWhenDone()` rather than an abrupt kill.

Confirmed empirically by the multi-method classes, which are exactly what would break on reuse-after-close. Note `OAuth2ResourceTest` is `abstract`, so its three concrete subclasses were run instead:

| class | passed |
|---|---|
| `OAuth2ResourceErrorForwardingTest` | 1/1 |
| `OAuth2ResourceRpIniLogoutTest` | 1/1 |
| `InMemSMOAuth2R2Test` | 7/7 |
| `JwtSMOAuth2R2Test` | 9/9 |
| `SyncSMOAuth2R2Test` | 7/7 |
| `InMemB2CResourceTest` | 23/23 |

## Timing

| | before | after | delta |
|---|---|---|---|
| `InMemB2CResourceIntegrationTest` (mean of 4) | 3375 ms | 3307 ms | −68 ms (−2.0%) |
| `InMemB2CResourceTest` (23 tests) | 7098 ms | 6475 ms | −623 ms (−8.8%) |

Essentially flat on the integration test, as that run is dominated by fixed costs this change cannot touch — the deliberate `expiresIn = 1`, router start/stop, and the 500-worker burst with its retry sleeps. The saving shows up where many sequential requests span many test methods. The justification for this change is resource hygiene, not wall-clock.

## On #3238

`SyncB2CResourceIntegrationTest` was run twice, purely to scan for new exception types, and passed both times. That is **not** evidence that #3238 is fixed — at a ~50% flake rate two consecutive passes are well inside noise (p≈0.25). The only signal taken from it is the negative one: no new exception type there either.

predic8 (dismissed):

predic8 (approved):

rrayst:
Re-verified on Windows 11 after the rebase onto current `master`, at `28f42dbf2`. This supersedes the results above, which were measured before #3237 was squash-merged and before master picked up the HTTP-client refactor.

`core` compiles against master's refactored `HttpClient` / `ConnectionFactory` / `RetryHandler` / protocol-handler API — fresh `mvn install -DskipTests -pl annot,core,test -am` then `test-compile`, both **BUILD SUCCESS**.

## Test battery — 54 test executions, all pass

| test | runs | found | passed | duration |
|---|---|---|---|---|
| `InMemB2CResourceIntegrationTest` | 4 | 1 each | **4/4** | 3067–3461 ms |
| `JwtB2CResourceIntegrationTest` | 2 | 1 each | **2/2** | 3194, 3212 ms |
| `InMemB2CResourceTest` | 1 | 23 | **23/23** | 6242 ms |
| `InMemSMOAuth2R2Test` | 1 | 7 | **7/7** | 1415 ms |
| `JwtSMOAuth2R2Test` | 1 | 9 | **9/9** | 3998 ms |
| `SyncSMOAuth2R2Test` | 1 | 7 | **7/7** | 1385 ms |
| `OAuth2ResourceErrorForwardingTest` | 1 | 1 | **1/1** | 743 ms |
| `OAuth2ResourceRpIniLogoutTest` | 1 | 1 | **1/1** | 1334 ms |

Zero throwables and zero WARN/ERROR lines throughout. `OAuth2ResourceTest` is `abstract`, so it is covered via its three concrete subclasses. Timing is unchanged from the pre-rebase run, within noise.

## Separately: the refactor did not break #3237

Worth recording here since it is on the path this PR sits on. #3237's fix depends on `HttpClient.call` propagating a **raw** `java.net.ConnectException` for `ConnectRetry` to catch, and master's refactor put a protocol-handler indirection in that path. It still works.

`MassivelyParallelTest` under the DEBUG config — the reproducer that failed 8/8 before #3237:

| run | result | `ConnectException` | reached 200 | total attempts |
|---|---|---|---|---|
| 1 | pass | 2070 | 500 | 2570 |
| 2 | pass | 1841 | 500 | 2341 |
| 3 | pass | 1759 | 500 | 2259 |
| 4 | pass | 2164 | 500 | 2664 |
| 5 | pass | 1953 | 500 | 2453 |

**5/5 pass**, with `attempts = refusals + 500` exactly in every run, so nothing is being lost. `ConnectRetry` is visible in the stack trace, which is the direct proof the exception arrives unwrapped:

```
java.net.ConnectException: Connection refused: getsockopt
	at com.predic8.membrane.core.transport.http.ConnectionFactory.getConnection(ConnectionFactory.java:84)
	at com.predic8.membrane.core.transport.http.HttpClient.dispatchCall(HttpClient.java:80)
	at com.predic8.membrane.core.transport.http.client.RetryHandler.executeWithRetries(RetryHandler.java:122)
	at com.predic8.membrane.core.transport.http.HttpClient.call(HttpClient.java:72)
	at com.predic8.membrane.core.transport.http.ConnectRetry.call(ConnectRetry.java:44)
	at com.predic8.membrane.integration.withoutinternet.MassivelyParallelTest.parallelTestWorker(MassivelyParallelTest.java:131)
```

The attempt distribution is identical to pre-refactor: 203–207 requests succeed on the first attempt (the accept queue Windows caps at 200), the rest take 3–12, worst case 12 of the 50 allowed.

One incidental note on master's refactor, benign and not part of this PR: `HttpClient.call` invokes `ph.cleanup(exc)` after `executeWithRetries` without a `try/finally`, so cleanup is skipped when a connect is refused — which `ConnectRetry` now causes up to 12 times per request. It is a no-op in that case, because `Http1ProtocolHandler.cleanup` returns immediately when `exchange.getTargetConnection() == null`, which is exactly the refused-connect state. Flagged only because it sits on the path the retry depends on.

PR #3266, opened 2026-09-13
fix(http): reassemble XOP in getBodyAsStringDecoded

### LINKED ISSUE #3126: getBodyAsStringDecoded() does not reassemble XOP/MTOM, unlike getBodyAsStreamDecoded()
`Message` has two "decoded" body accessors that decode differently: `getBodyAsStreamDecoded()` reassembles XOP/MTOM multiparts, `getBodyAsStringDecoded()` does not. The names and Javadoc suggest they are the same operation in two shapes, so callers pick whichever return type is convenient and silently get different content for an MTOM message.

### Evidence

`getBodyAsStreamDecoded()` (`core/src/main/java/com/predic8/membrane/core/http/Message.java:136`) reconstitutes XOP first, then falls back to Content-Encoding handling:

```java
Message m = xopr.getReconstitutedMessage(this);
if (m != null)
    return m.getBodyAsStream(); // we know decoding is not necessary any more
return MessageUtil.getContentAsStream(this);
```

`getBodyAsStringDecoded()` (`Message.java:164`) goes straight to `MessageUtil.getContent(this)`:

```java
return new String(MessageUtil.getContent(this), getCharsetOrDefault());
```

and `MessageUtil.getContent` (`core/src/main/java/com/predic8/membrane/core/util/MessageUtil.java:52`) only handles gzip/deflate/brotli before returning `msg.getBody().getContent()` — no XOP step.

So for a `multipart/related; type="application/xop+xml"` message:

| accessor | result |
| --- | --- |
| `getBodyAsStreamDecoded()` | reassembled SOAP envelope, `xop:Include` replaced by base64 |
| `getBodyAsStringDecoded()` | the raw multipart package, MIME boundaries and all |

### Why it matters

The two are documented as interchangeable — `getBodyAsStringDecoded()`'s Javadoc says "Use `getBodyAsStreamDecoded()` instead" purely on performance grounds, implying the content is the same. There are ~43 files calling `getBodyAsStringDecoded()`; anything reached by an MTOM message gets the multipart wrapper instead of the envelope. Failures are silent and content-dependent: an XPath or a `contains` check just stops matching, rather than erroring.

Same split exists one level down and is worth settling at the same time: `MessageUtil.getContentAsStream` vs `MessageUtil.getContent` are an encoding-only pair, while only `Message.getBodyAsStreamDecoded` adds XOP.

### Possible resolutions

1. Make `getBodyAsStringDecoded()` reconstitute too, so "Decoded" means one thing. Behaviour change for MTOM traffic, but the direction the names already promise.
2. Leave behaviour alone and rename/document so the difference is explicit (e.g. keep XOP out of both and give reconstitution its own named accessor).

Either is fine; the current state — same suffix, different semantics, no note anywhere — is the problem.

Found while working on `XmlDomBody` (WS-Security branch), which parses from `getBodyAsStreamDecoded()`. Not blocking that work.

arimu1:
## Summary

- `getBodyAsStringDecoded()` now reads bytes from `getBodyAsStreamDecoded()`, so MTOM/XOP messages return the reassembled SOAP envelope instead of the raw multipart package.
- Adds a regression test ensuring string and stream decoded accessors agree on synthetic XOP fixtures.

Fixes #3126

## Test plan

- [x] `test/scripts/run-core-test.sh com.predic8.membrane.core.multipart.XOPReconstitutorTest` (14/14 passed, JDK 21)


predic8:
@arimu1 are you a human? 250 PRs in two weeks is a lot

PR #3276, opened 2026-09-14
fix(core): complete the HTTP status code message table

### LINKED ISSUE #3275: HttpUtil.getMessageForStatusCode returns "" for unlisted codes, so 429 and others ship a blank reason phrase
`HttpUtil.getMessageForStatusCode` maps ~25 status codes and returns `""` for everything else:

```java
// core/src/main/java/com/predic8/membrane/core/util/HttpUtil.java:177
public static String getMessageForStatusCode(int code) {
    return switch (code) {
        case 200 -> "OK";
        ...
        case 504 -> "Gateway Timeout";
        default -> "";
    };
}
```

Missing are a number of codes Membrane itself emits or lets users configure, most notably
**429**, plus 203, 205, 207, 402, 406, 407, 408, 410, 411, 412, 413, 414, 416, 417, 421, 423,
425, 426, 428, 431, 451, 505, 507 and 511. (205 is a nice illustration: `shouldNotContainBody()`
in `Response.java:462` knows about it, the message table does not.)

## Impact

`ResponseBuilder.status(int)` (`Response.java:69-72`) sets the status message from this method,
so an unlisted code ends up with an empty reason phrase, which `Response.getStartLine()`
(`Response.java:356`) interpolates verbatim:

```
HTTP/1.1 429 
```

This is reachable in shipped code. `RateLimitInterceptor:116` builds its response with
`ProblemDetails.user(...).status(429)`, and that path runs
`ProblemDetails.createContent` → `Response.statusCode(429)` → `ResponseBuilder.status(429)` →
`getMessageForStatusCode(429)` → `""`. **Every response the rate limiter rejects a client with
carries a blank reason phrase.** The same applies to any user config that picks an unlisted
code, e.g. `<return statusCode="429"/>` (`ReturnInterceptor:89`).

Second, more visible symptom: `Response.fromStatusCode` (`Response.java:333-338`) builds an HTML
error page whose title is `"%d %s.".formatted(statusCode, getMessageForStatusCode(statusCode))`.
For an unlisted code that renders as a title of literally `429 .` — number, space, full stop.

## Severity

To be precise: an empty reason phrase is **not** a protocol violation. RFC 9112 §4 defines
`status-line = HTTP-version SP status-code SP [ reason-phrase ]`, the phrase is optional, and
clients are told to ignore it. So this does not break interoperability.

It does hurt in two smaller ways: the HTML error page above is visibly broken, and anything that
surfaces the reason phrase — `curl -i`, access logs, proxies, debugging tools — shows nothing
where it should show `Too Many Requests`. For a gateway whose job is to produce other people's
error responses, that seems worth getting right.

## Suggested fix

Complete the switch from the IANA HTTP Status Code Registry, and make `default` fall back to a
generic phrase derived from the status class (e.g. `"Client Error"` / `"Server Error"`) rather
than `""`, so a code nobody anticipated still produces a well-formed status line.

Worth a test that asserts every code the switch lists round-trips, and that an unregistered code
(say 499) yields a non-empty phrase.

---
Found while reviewing the `status` attribute added to `wsdl2openapi`'s `operationSettings`, which
also routes through `getMessageForStatusCode`. Filing separately since the defect is in
`HttpUtil` and predates that work.

wangzhengzhuo05:
Fixes #3275.

## What

`HttpUtil.getMessageForStatusCode` returned `""` for every status code outside its ~25-entry
table, so the status line and the generated HTML error page came out malformed for codes the
gateway itself emits or lets users configure.

`ResponseBuilder.status(int)` (`Response.java:69-72`) takes the reason phrase straight from this
method, and `Response.getStartLine()` interpolates it verbatim, so a code like 429 shipped as:

```
HTTP/1.1 429
```

That path is reachable in shipped code — `RateLimitInterceptor` rejects clients with
`ProblemDetails.user(...).status(429)` — and in user config via `<return statusCode="429"/>`
(`ReturnInterceptor:89`). The second symptom is `Response.fromStatusCode` (`Response.java:333-338`),
which titles the HTML error page `"%d %s."` and so rendered literally `429 .`.

## How

1. Completed the switch from the IANA HTTP Status Code Registry: added 101, 102, 103, 203, 205,
   207, 208, 226, 300, 305, 402, 406, 407, 408, 410, 411, 412, 413, 414, 416, 417, 421, 423, 424,
   425, 426, 428, 429, 431, 451, 505, 506, 507, 508, 510, 511. Pre-existing entries are unchanged,
   including the non-standard `415 -> "Unsupported Mediatype"` wording.
2. `default` now delegates to a private `defaultMessageForStatusCode(int)` that derives a phrase
   from the status class (`Information` / `Success` / `Redirection` / `Client Error` /
   `Server Error`), so a code nobody anticipated still produces a well-formed status line instead
   of an empty phrase.

No public signature changed and no new dependency was added.

## Tests

Added to the existing `HttpUtilTest` (mirrored test class for `HttpUtil`):

- `statusMessageForListedCode` — parameterized, asserts the reason phrase for a representative set
  of the newly added codes (205, 207, 413, **429**, 431, 451, 505, 511) plus 204/301.
- `unlistedCodeFallsBackToItsStatusClass` — parameterized, asserts 199/299/399/499/599 yield the
  class-based phrase.
- `everyStatusCodeHasANonEmptyMessage` — sweeps 100..599 and asserts no blank reason phrase.

```
./test/scripts/run-core-test.sh com.predic8.membrane.core.util.HttpUtilTest
44 tests found, 44 tests successful, 0 tests failed
```
(28 passed before the new tests were added, on the same checkout.)

## Mutation check

Both halves of the fix were verified by reverting them one at a time:

- `default -> defaultMessageForStatusCode(code)` → `default -> ""` ⇒ **6 failures**
  (`everyStatusCodeHasANonEmptyMessage` + all five `unlistedCodeFallsBackToItsStatusClass` cases).
- deleting `case 429 -> "Too Many Requests";` ⇒ **1 failure**, exactly
  `statusMessageForListedCode[6] 429, Too Many Requests`.

After restoring: 44/44 green.

## AI-generated code disclosure

The change was implemented with AI assistance (OpenCode with muse-spark 1.3) and reviewed and
verified locally for correctness (build, single-class test run, mutation checks above) before
submission.

## Summary by CodeRabbit

- **Improvements**
  - HTTP responses now display standardized reason phrases for a broader range of status codes.
  - Previously unrecognized status codes now show a meaningful category such as “Information,” “Success,” “Redirection,” “Client Error,” or “Server Error” instead of a blank message.

- **Tests**
  - Added comprehensive coverage for standard, uncommon, and fallback HTTP status messages, including validation that all status codes from 100 through 599 return a message.


predic8:
Explain me 226

wangzhengzhuo05:
`226 IM Used` is IANA-registered — RFC 3229 (Delta Encoding in HTTP). A server sends it when it answers a `GET` with a delta against a representation the client already holds, and the instance manipulations ("IM") were applied while producing it. It is the only registered 2xx code outside the 200–208 run, which is why it looks out of place between `208` and `300`.

While I was here I also picked up CodeRabbit's note and added `104 Upload Resumption Supported` — the one registered code that was still missing from the table (`cd9405950`). Registered-vs-not is the line I drew: `418` is not in the IANA registry and `306` is marked unused/reserved, so both are left out and keep falling through to the class-based phrase instead.

`HttpUtilTest` is 46/46 green locally (`test/scripts/run-core-test.sh com.predic8.membrane.core.util.HttpUtilTest`). The full CI on this PR is still waiting on a `/ok-to-test` from an authorized committer — happy to fix anything it flags once it runs.

predic8:
@wangzhengzhuo05 we do not want to bloat the code. But you can add just 406-408, 412, 418, 423, 428, 429, 508

predic8:
@wangzhengzhuo05 are you finishing this?

christiangoerdes:
superseded by #3298

PR #3281, opened 2026-09-15
Fix flaky plaintext assertion in EncryptSecurePartTest

christiangoerdes:
There is a chance (about 0.17%) that "bar" is part of the Base64 encoding. 

## Summary by CodeRabbit

* **Tests**
  * Strengthened validation of encrypted SOAP messages to confirm body content is properly protected and remains decryptable.


core/src/test/java/com/predic8/membrane/core/interceptor/soap/wsse/EncryptSecurePartTest.java
```diff
@@ -97,7 +97,7 @@ void encryptsTheBodyContentAndLeavesTheBodyElementInPlace() throws Exception {
         Element body = firstByTag(doc, SOAP_NS, "Body");
         assertEquals(1, body.getElementsByTagNameNS(XENC_NS, "EncryptedData").getLength());
         assertEquals(0, doc.getElementsByTagName("foo").getLength(), "the plaintext must be gone");
-        assertFalse(rawBody().contains("bar"), "the plaintext must not survive anywhere in the message");
+        assertFalse(rawBody().contains(">bar<"), "the plaintext must not survive anywhere in the message");
```
predic8: Let's use something longer than bar as plaintext e.g.

Cheesesteak

predic8 (approved):

PR #3317, opened 2026-09-20
Fix flaky ScriptingGroovyTutorialTest.groovyEndpointLogs

predic8:
## Summary
- `groovyEndpointLogs` captured `System.out` and asserted on it immediately after the HTTP call returned, but the gateway under test runs as a separate process whose console output is relayed into the test JVM asynchronously by a background reader thread (`Process2`'s `OutputWatcher`).
- The RESPONSE-flow log line is emitted late in the exchange lifecycle and can still be in flight (not yet relayed) when the response comes back over HTTP, causing an intermittent `assertTrue` failure on that line while the REQUEST-flow line (emitted earlier, with more margin) passes.
- Fix: poll the captured output for up to 10×100ms instead of checking once, mirroring the existing polling pattern used in `HotReloadTutorialTest`.

## Test plan
- [x] Ran `com.predic8.membrane.tutorials.advanced.ScriptingGroovyTutorialTest` 3x in a row against a freshly built distribution zip — 3/3 tests passed each run, 0 failures.

## Summary by CodeRabbit

* **Tests**
  * Improved reliability when verifying asynchronous request- and response-flow logging.
  * Added event-based waiting with a defined timeout to reduce intermittent failures and provide clearer test outcomes.
  * Ensured monitoring is cleaned up after each test, including when an assertion fails.
  * Confirmed successful request processing before validating the expected gateway log messages.


rrayst:
Monitoring the process's should not be done via swapping `System.out`, but by using `Process2.Builder.withWatcher(...)`. A watcher gets pushed the process's stdout and stderr output. Use a `new CountDownLatch(1)` within the watcher, call `.countDown()` on it, once the expected output arrives, and call `.await()` on it from the test. This eliminates the poll loop.

rrayst (approved):

